PR #1, opened 2020-07-22
Adding a CODEOWNERS file.

Ratchette:


CODEOWNERS
```diff
@@ -0,0 +1,6 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# @global-owner1 and @global-owner2 will be requested for
```
allie-wood, line 3: So does this mean it would actually request both of us as reviewers for any pull request?

allie-wood (approved):

CODEOWNERS
Ratchette: Yep! This means that we will both be added as reviewers. 
As a follow-up to this change I'll configure our github workflow so that only one approval is neccessary.

Ratchette: https://docs.github.com/en/github/administering-a-repository/enabling-required-reviews-for-pull-requests

PR #5, opened 2020-07-29
Added protocol buffer models 

margaritiko:
Add protocol buffer models for representing __bindings__, __dependencies__ and __graph__

All models constructed with `proto2`. This is preferred because in `proto3` all fields are optional which can add ambiguity.

- `Binding` is represented by its unique string-key.
- `Dependency` model represents an edge in graph which always contains target node.

- `BindingGraph` contains a map `adjacency_list` which simply matches binding or component (access it by string representation of key) to list with other bindings.


project/src/com/google/daggerquery/protobuf/autogen/BindingGraphProto.java
```diff
@@ -0,0 +1,1932 @@
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
```
margaritiko: All these files with `Proto` suffix are generated automatically, so don't waste your time on them ⏰

project/src/com/google/daggerquery/protobuf/binding.proto
```diff
@@ -0,0 +1,52 @@
...
+    MULTIBOUND_SET = 10;
+    MULTIBOUND_MAP = 11;
+    OPTIONAL = 12;
+    DELEGATE = 13;
+    MEMBERS_INJECTION = 14;
+  }
+
+  required string component_name = 1;
+  required bool is_component     = 2;
+  required bool is_missing       = 3;
```
margaritiko, line 45: Two fields `is_component` and `is_missing` needed for distinguishing three different classes: [BindingNode](https://github.com/google/dagger/blob/4a3115780c837f0857e9a14b09ee76a932d8dd80/java/dagger/internal/codegen/binding/BindingNode.java#L45), [MissingBinding](https://github.com/google/dagger/blob/4a3115780c837f0857e9a14b09ee76a932d8dd80/java/dagger/model/BindingGraph.java#L406) and [ComponentNode](https://github.com/google/dagger/blob/4a3115780c837f0857e9a14b09ee76a932d8dd80/java/dagger/model/BindingGraph.java#L443).

project/src/com/google/daggerquery/protobuf/autogen/BindingGraphProto.java
kwchan1995: How about excluding generated code from source control, by marking them in `.gitignore`? And maybe add a command for the code generation in the `README`. This way it will reduce clutter and prevent the generated / actual source code from ever going out of sync :)

kwchan1995:
Perhaps controversial, actually the internal guidance is to prefer `optional` protobuf fields to `required`, because it makes future refactorings easier without having to support the `required` fields indefinitely (see https://developers.google.com/protocol-buffers/docs/proto#specifying_field_rules and https://github.com/protocolbuffers/protobuf/issues/2497#issuecomment-267422550). This is also why `required` was removed in `proto3`. Would it be possible to move to `proto3` for future-proofing if not too much work please?

project/src/com/google/daggerquery/protobuf/binding_graph.proto
```diff
@@ -0,0 +1,33 @@
...
+option java_package = "com.google.daggerquery.protobuf";
+option java_outer_classname = "BindingGraphProto";
+
+message BindingGraph {
+
+  message ListWithDependencies {
+    repeated Dependency dependency = 1;
+  }
+
+  map<string, ListWithDependencies> adjacency_list = 1;
```
kwchan1995, line 38: The direction of the edges are actually important, 

```
class A {
  @Inject A(B b) { ... }
}
```

is different from

```
class B {
  @Inject B(A a) { ... }
}
```

You might need to adapt the data structure to ensure the relationship is preserved.

Could you please add inline documentation for each class member (could be a short sentence, just so to avoid ambiguity of interpretation by the reader)? Thanks!

project/src/com/google/daggerquery/protobuf/autogen/BindingGraphProto.java
margaritiko: Cool, thanks! I think it's good idea, I will add entire `autogen` folder into .gitignore

project/src/com/google/daggerquery/protobuf/binding_graph.proto
margaritiko: I think this data structure provides directions 🤔

I mean with your example we will have adjacency_list[`A`] = [..., `B`, ...] for edge A->B (and `A` might not be presented in appropriate list for `B` if there is no edge B->A).

Yes, I will provide documentation in next commit, sure!

kwchan1995: Okay, thanks for the clarification! I was initially confused because the https://en.wikipedia.org/wiki/Adjacency_list page defined it for undirected graphs, but your definition would work as well. Please do add a comment (in the next commit is fine) to prevent people like me coming to bug you in the future :)

README.md
```diff
@@ -4,6 +4,17 @@ A graph based query language on top of [dagger](https://github.com/google/dagger
 
 Dagger Query can be used to investigate the dagger dependency graph.
 
+## Before you start
```
Ratchette, line 7: Adding this documentation to the readme is a really good idea.

Ratchette (approved):

PR #6, opened 2020-08-03
Implemented an algorithm for traversing binding graph

margaritiko:
Implement __Depth First Search__ algorithm to convert `BindingGraph` instance into adjacency list.

### Main changes:
* Put all logic of converting graph into special generic type `GraphConverter` which is responsible for making adjacency list and provides one method `makeBindingGraphProto(NodeT rootNode, Network<NodeT, EdgeT> network)` for it.

* Added dependencies for `Guava` library for using package `com.google.common.graph` and for `JUnit` library to support tests.

* Covered with unit tests logic of `GraphConverter` class.

[Card №1](https://github.com/googleinterns/dagger-query/projects/1#card-42487416)
[Card №2](https://github.com/googleinterns/dagger-query/projects/1#card-42487436)


project/src/com/google/daggerquery/plugin/GraphConverter.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class GraphConverter<NodeT, EdgeT> {
+  /**
+   * Makes a directed adjacency list based on a given {@link com.google.common.graph.Network} instance
+   * and writes it to a proto model.
+   *
+   * <p>Uses models which were generated with <b>protocol buffers library</b> for representing edges.
```
Ratchette: Would it be possible to include a link here?
https://developers.google.com/protocol-buffers

project/src/com/google/daggerquery/plugin/GraphConverter.java
```diff
@@ -0,0 +1,99 @@
...
+   *
+   * <p>Saving the directions of edges is provided by data structure itself.
+   * For example, if we have an edge A --> B, we can be sure that list
+   * will look like this: {@code adjacencyList["A"] = {..., "B", ...}}.
+   *
+   * @see <a href="https://en.wikipedia.org/wiki/Adjacency_list">Adjacency list description</a>
+   */
+  BindingGraphProto.BindingGraph makeBindingGraphProto(NodeT rootNode, Network<NodeT, EdgeT> network) {
+    Map<String, ListWithDependencies> adjacencyList = new HashMap<>();
+    Set<String> visitedNodes = new HashSet<>();
```
Ratchette, line 58: It looks like this member is only used int he method below.
Can you please move this variable to be defined in that method instead of passed as a param?

Ratchette: Ignore this - I didn't read far enough to realize that "makeAdjacencyList" is recursive.

project/tests/com/google/daggerquery/plugin/GraphConverterTest.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.daggerquery.protobuf.autogen.BindingGraphProto;
+import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph.ListWithDependencies;
+import org.junit.Assert;
+import org.junit.Test;
+import java.util.Map;
+
+public class GraphConverterTest {
+  @Test
+  public void testMakingGraph_WhenContainsOneEdge() {
+    MutableNetwork<String, Integer> network = makeMutableDirectedNetwork(/*allowsParallelEdges = */ false);
```
Ratchette, line 31: This is really good practice :)

Ratchette (approved):
Looks good to me.
You may want to wait until Allie or Karl (aka someone more familiar with dagger's binding graph) to take a look too.

Some general notes:
- The amount of comments that you put in (and the amount of detail that you put into each comment) is wonderful
-  The comment that you put inside each call to makeMutableDirectedNetwork is also very helpful.
- You also did a great job of dividing the code into small, bite-sized methods.

project/src/com/google/daggerquery/plugin/GraphConverter.java
margaritiko: ✅✅✅

project/BUILD
```diff
@@ -23,13 +23,33 @@ java_library(
     deps = ["//third_party/java/protobuf:protobuf"],
 )
 
-java_plugin(
-    name = "dagger_query_plugin",
-    srcs = ["src/com/google/daggerquery/plugin/QueryPlugin.java"],
+java_library(
+    name = "dagger_query_plugin_sources",
+    srcs = glob(["src/com/google/daggerquery/plugin/**/*.java"]),
```
kwchan1995: The convention in Google is to have a BUILD file in each folder that has sources (https://docs.bazel.build/versions/master/build-ref.html#packages), so that you never have to make cross-folder references in srcs. This results in better modularity.

So in your case it would be something like:
src/com/google/daggerquery/BUILD
src/com/google/daggerquery/plugin/BUILD
tests/com/google/daggerquery/BUILD
tests/com/google/daggerquery/plugin/BUILD
etc.

project/BUILD
```diff
@@ -23,13 +23,33 @@ java_library(
...
         ":dagger_query_protobuf",
         "//third_party/java/auto:service",
+        "//third_party/java/guava:guava",
+    ]
+)
+
+java_test(
+    name = "DaggerQueryTest",
+    size = "small",
+    srcs = glob(["tests/com/google/daggerquery/**/*.java"]),
```
kwchan1995: Please create a different test target per test (i.e. don't use glob for tests). This allows tests to be run standalone.

This way you don't need to specify the test_class attribute, because the srcs will only contain a single test and is unambiguous.

project/src/com/google/daggerquery/plugin/GraphConverter.java
```diff
@@ -0,0 +1,100 @@
...
+import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph.ListWithDependencies;
+import com.google.daggerquery.protobuf.autogen.DependencyProto.Dependency;
+import dagger.model.Binding;
+import dagger.model.BindingGraph;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class GraphConverter<NodeT, EdgeT> {
```
kwchan1995, line 42: Since the GraphConverter is specialised for Dagger, how about setting upper bounds for NodeT and EdgeT to https://dagger.dev/api/2.23.1/dagger/model/BindingGraph.Node.html and https://dagger.dev/api/2.23.1/dagger/model/BindingGraph.Edge.html respectively for better type safety?

margaritiko: Yes, it makes sense to add this requirements for types 🖊

But what about tests? Now I'm using `Strings` as Nodes and as there are no external implementations of `BindingGraph.Node` it could be a problem.

kwchan1995: Never mind, I just realised you had to pass generic types for testing, which makes sense.

project/tests/com/google/daggerquery/plugin/GraphConverterTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    Map<String, ListWithDependencies> adjacencyList = makeAdjacencyList(/*rootNode = */ "A", network);
+
+    Assert.assertEquals(0, adjacencyList.get("C").getDependencyCount());
+    Assert.assertEquals(1, adjacencyList.get("B").getDependencyCount());
+    Assert.assertEquals(2, adjacencyList.get("A").getDependencyCount());
+  }
+
+  @Test
+  public void testMakingGraph_WhenContainsParallelEdges() {
```
kwchan1995, line 57: It's great that you thought of this case! :) 

(In reality this shouldn't happen in the Dagger graph.)

project/src/com/google/daggerquery/plugin/GraphConverter.java
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Implementation of depth first search algorithm that traverses a given network.
+   *
+   * <p>For each node, all its child nodes are saved to {@code adjacencyList}.
+   *
+   * @param currentNode is a node which will be processed at this step
+   * @param visitedNodes is a set that contains all processed nodes to avoid loops
+   */
+  private void makeAdjacencyList(NodeT currentNode, Network<NodeT, EdgeT> network,
```
allie-wood, line 83: Could you maybe add a comment here specifying what the outcome of calling this method is? For example saying it will populate adjacencyList I suppose?

project/tests/com/google/daggerquery/plugin/GraphConverterTest.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.common.graph.MutableNetwork;
+import com.google.common.graph.Network;
+import com.google.common.graph.NetworkBuilder;
+import com.google.daggerquery.protobuf.autogen.BindingGraphProto;
+import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph.ListWithDependencies;
+import org.junit.Assert;
+import org.junit.Test;
+import java.util.Map;
+
+public class GraphConverterTest {
```
allie-wood, line 32: Do you think it might be worth putting a couple more test cases here, such as what happens when the network doesn't contain anything? Or calling the methods with some other bad input. Maybe that couldn't happen, I'm not sure?

project/src/com/google/daggerquery/plugin/GraphConverter.java
allie-wood, line 42: Could you add a javadoc comment on the class, maybe specifying what NodeT and EdgeT are?

project/tests/com/google/daggerquery/plugin/GraphConverterTest.java
```diff
@@ -0,0 +1,76 @@
...
+  @Test
+  public void testMakingGraph_WhenContainsOneEdge() {
+    MutableNetwork<String, Integer> network = makeMutableDirectedNetwork(/*allowsParallelEdges = */ false);
+    network.addEdge("A", "B", 10);
+
+    Map<String, ListWithDependencies> adjacencyList = makeAdjacencyList(/*rootNode = */ "A", network);
+
+    Assert.assertTrue(adjacencyList.get("A").getDependency(0).getTarget().equals("B"));
```
allie-wood: Optional nit: you could static import Assert.assertTrue so that you don't have to repeat it

allie-wood (approved):
Looking great! 👍

project/BUILD
margaritiko: Done ✅

However it shows an error `cannot determine test class` and for some reason class name cannot be inferred using the target's name and its source-root-relative path, so I added `test_class`.

project/src/com/google/daggerquery/plugin/GraphConverter.java
margaritiko: Done ✅

project/src/com/google/daggerquery/plugin/GraphConverter.java
margaritiko: Great idea, thanks, I've added it ✔️

project/tests/com/google/daggerquery/plugin/GraphConverterTest.java
margaritiko: Yeah, since method `makeAdjacencyList` is public it makes sense to show what will happen in worse situation!

I've added tests with passing null arguments, with passing empty network (it's also incorrect because we assume that in binding graph we have at least one node - main component) and with passing invalid source node.

project/tests/com/google/daggerquery/plugin/GraphConverterTest.java
margaritiko: Thanks, I've added it

project/BUILD
kwchan1995: Thanks!

Interesting, I'm not sure why the `cannot determine test class` error pops up, as the [docs](https://docs.bazel.build/versions/master/be/java.html#java_test.test_class) suggest the `test_class` will be inferred from the target's `name` and its source-root-relative path when omitted. In any case this is a really minor nit, so let's not get bogged down. :smiley:

kwchan1995:
Looks great to me!

PR #8, opened 2020-08-05
Added presubmit.yml file required for opening CI issue

margaritiko:
Create `presubmit.yml` file with three supported platforms required for opening CI issue

> This changes made in order to open an issue for adding CI to this project (as described in [README](https://github.com/bazelbuild/continuous-integration))

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,0 +1,24 @@
+---
+bazel: 2.1.0
+platforms:
+  ubuntu1804:
+    run_targets:
+    - "..."
+    build_targets:
+    - "..."
+    test_targets:
+    - "//tests/com/google/daggerquery/plugin:GraphConverterTest"
+  macos:
+    run_targets:
+    - "..."
+    build_targets:
+    - "..."
+    test_targets:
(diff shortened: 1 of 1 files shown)


Ratchette (approved):
We might want to add build targets in the future too.

.bazelci/presubmit.yml
```diff
@@ -0,0 +1,24 @@
+---
+bazel: 2.1.0
+platforms:
+  ubuntu1804:
+    run_targets:
+    - "..."
+    build_targets:
+    - "//src/com/google/daggerquery:dagger_query_plugin"
+    test_targets:
+    - "//tests/com/google/daggerquery/plugin:GraphConverterTest"
```
kwchan1995: You might want to consider using a wildcard suffix, e.g. `"//tests/..."` so that you don't have to maintain this file every time a new target is added to the codebase.

kwchan1995 (approved):
Thanks for taking the time to set up continuous integration!

.bazelci/presubmit.yml
margaritiko: Done, thanks!

PR #9, opened 2020-08-06
Specified working directory and removed bazel version constraint

margaritiko:
Fixes #8 with specifying working directory and removing `bazel` version

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,5 +1,4 @@
 ---
-bazel: 2.1.0
 platforms:
   ubuntu1804:
     run_targets:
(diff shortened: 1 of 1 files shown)


Ratchette (approved):
LGTM 🐳

Copy&pasting more context from an email thread:
* ["working_directory" should point to the directory that contains bazel's WORKSPACE file.](https://github.com/bazelbuild/continuous-integration/tree/master/buildkite#using-multiple-workspaces-in-a-single-pipeline)
* [Removing the "bazel_version" field it will automatically fetch the latest Bazel release (this is equal to 'bazel_version: latest')](https://github.com/bazelbuild/continuous-integration/tree/master/buildkite#using-a-specific-version-of-bazel)

kwchan1995 (approved):

PR #10, opened 2020-08-07
Added bazel rule with example to test graph serialization

margaritiko:
Add bazel rule to extract `.textproto` file from app jar

Now all process looks like this:
* In dagger spi plugin with `Filer` instance we save a file with binding graph to `binding_graph.textproto`
* This file with graph appears in main app's jar
* With rule `dagger_query_textproto` we extract file from this jar and now able to add it to any other target 🎉

### DIFF
--- a/project/src/com/google/daggerquery/dagger_query_textproto.bzl
+++ b/project/src/com/google/daggerquery/dagger_query_textproto.bzl
@@ -1,0 +1,42 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+def dagger_query_textproto(name, dagger_app_target):
+    _dagger_query_textproto(name = name, dagger_app_target = dagger_app_target)
+
+def _dagger_query_textproto_impl(ctx):
+    dagger_app_target = ctx.attr.dagger_app_target
+    src_jars = dagger_app_target[JavaInfo].source_jars
+
+    ctx.actions.run_shell(
+        inputs = src_jars,
+        outputs = [ctx.outputs.out],
+        command = "unzip -p {src_jar} binding_graph.textproto > {out}".format(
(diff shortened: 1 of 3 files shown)


project/src/com/google/daggerquery/dagger_query_textproto.bzl
```diff
@@ -0,0 +1,28 @@
...
+
+def _dagger_query_textproto_impl(ctx):
+    dagger_app_target = ctx.attr.dagger_app_target
+    src_jars = dagger_app_target[JavaInfo].source_jars
+
+    ctx.actions.run_shell(
+        inputs = src_jars,
+        outputs = [ctx.outputs.out],
+        command = "unzip -p -o {src_jar} binding_graph.textproto > {out}".format(
+            src_jar = src_jars[0].path,
```
kwchan1995, line 29: Sorry Rita I ran into problems setting this up internally so I wasn't able to debug this. I think what you can do is to print out the path of this `src_jar`, and manually examine the contents of this jar to ensure the `binding_graph.textproto` file is actually there. This will help narrow down whether it's a problem with the shell command, we're feeding in the wrong jar, or there was a problem packaging the textproto file into the jar.

margaritiko: No problem, it's good idea, thanks!
I will try to reproduce these steps 👌

margaritiko:
@Ratchette @kwchan1995 @allie-wood  
Finally I solved this problem and now rule is correct.
I've removed all example sources and PR became smaller 😊

I also updated description of this PR with up-to-date information about process of file's generation ✨

project/src/com/google/daggerquery/dagger_query_textproto.bzl
```diff
@@ -0,0 +1,42 @@
...
+
+def _dagger_query_textproto_impl(ctx):
+    dagger_app_target = ctx.attr.dagger_app_target
+    src_jars = dagger_app_target[JavaInfo].source_jars
+
+    ctx.actions.run_shell(
+        inputs = src_jars,
+        outputs = [ctx.outputs.out],
+        command = "unzip -p {src_jar} binding_graph.textproto > {out}".format(
+            src_jar = src_jars[0].path,
```
kwchan1995, line 29: May be worth adding an assert statement that `src_jars` is of length 1. This way it can fail earlier for easier debugging instead of the error silently propagated downstream.

kwchan1995 (approved):
Awesome!

Ratchette (approved):
LGTM 🦙

project/src/com/google/daggerquery/dagger_query_textproto.bzl
margaritiko: Done ✅

PR #11, opened 2020-08-10
Fixing CI with buildkite

margaritiko:
Fixing CI with buildkite

* Removed run targets because now only plugin exists which cannot be launched standalone
* Added autogenerated files with java proto

  **IMPORTANT:** Later these files will be removed and replaced with `java_proto_library` which generates java proto at compile time [[Card]](https://github.com/googleinterns/dagger-query/projects/1#card-43452911)

[The latest build](https://buildkite.com/bazel/dagger-query/builds/12)


margaritiko:
I am adding some changes in order to fix CI. It's more complicated than I though because I also need to add command for generation protocol buffers models.

I will update this PR with final changes and merge only after being sure that build is fixed 🌟

Ratchette (approved):

kwchan1995 (approved):

PR #12, opened 2020-08-10
Supported DEPS query

margaritiko:
Support `deps` query and cover all models and services with tests

**All changes in this PR can be separated in two parts:**
* _implementation_
* _tests_

For supporting queries from user new `java_binary` rule with script sources was added. It simply uses arguments provided in `QueryExecutor` as a query and tries to parse it.

**Added new classes, namely:**
* `Query` - a class for representing query and its parameters and for executing queries with all logic of traversing a graph
* `SourcesLoader` -a class which defines entity responsible for deserialization of a binding graph
* `QueryExecutor` - a class with `main` method which parses command line arguments, deserializes binding graph, executes query and writes output for user

[Card №1](https://github.com/googleinterns/dagger-query/projects/1#card-42487514)
[Card №2](https://github.com/googleinterns/dagger-query/projects/1#card-42487560)


project/src/com/google/daggerquery/executor/services/DaggerQueryController.java
```diff
@@ -0,0 +1,51 @@
...
+      switch (query.getType()) {
+        case DEPS:
+          List<String> dependencies = queryExecutor.executeDepsQuery(
+              query.getParameters()[0],
+              sourcesLoader.loadBindingGraph()
+          );
+          outputWriter.writeDepsOutput(dependencies);
+          break;
+      }
+    } catch (ArrayIndexOutOfBoundsException e) {
```
kwchan1995: The error messages in the catch blocks assume implementation details (nested) within the try block, which could be fragile. Consider defining the error messages at the throw site of the exception instead.

project/src/com/google/daggerquery/executor/services/InternalSourcesLoader.java
```diff
@@ -0,0 +1,37 @@
...
+import java.io.InputStream;
+
+public class InternalSourcesLoader implements SourcesLoader {
+  private static final String PATH_TO_BINDING_GRAPH = "/com/google/daggerquery/binding_graph_data.textproto";
+
+  public BindingGraph loadBindingGraph() throws IOException {
+    InputStream inputStream = InternalSourcesLoader.class.getResourceAsStream(PATH_TO_BINDING_GRAPH);
+
+    if (inputStream == null) {
+      throw new FileNotFoundException("File binding_graph_data.textproto is missed.");
```
kwchan1995: How about interpolating the `PATH_TO_BINDING_GRAPH` variable to be more robust against name changes? :stuck_out_tongue:

project/src/com/google/daggerquery/executor/services/QueryExecutor.java
```diff
@@ -0,0 +1,71 @@
...
+   */
+  public List<String> executeDepsQuery(String sourceNode, BindingGraph bindingGraph) {
+    if (!bindingGraph.getAdjacencyListMap().containsKey(sourceNode)) {
+      // Will check if this node is a valid node in passed binding graph or not.
+      DependencyProto.Dependency targetNode = DependencyProto.Dependency.newBuilder().setTarget(sourceNode).build();
+
+      boolean containsValue;
+      containsValue = bindingGraph.getAdjacencyListMap().values().stream()
+          .map(list -> list.getDependencyList().contains(targetNode))
+          .reduce(false, (a, b) -> a || b);
```
kwchan1995: Can be simplified with the [`Stream.anyMatch`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#anyMatch-java.util.function.Predicate-) operator.

Alternatively, how about changing the adjacency list representation, such that a leaf node is associated with an empty `ListWithDependencies` (as opposed to the leaf node being absent)? This way you don't have to loop through all the adjacency lists.

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  /**
+   * Writes an output of execution {@link com.google.daggerquery.executor.models.QueryType}.DEPS query.
+   *
+   * <p>An output is an array with all dependencies presented as Strings,
+   * so method simply writes them one by one.
+   */
+  public void writeDepsOutput(List<String> dependencies) {
+    printStream.println(String.format("All dependencies:"));
```
kwchan1995: How about making the output minimal and simply print a dep on each line?

This won't be the friendliest representation to the user, but it will simplify parsing for other command line tools that wish to reuse the output of this tool.

project/tests/com/google/daggerquery/executor/models/QueryTest.java
```diff
@@ -0,0 +1,69 @@
...
+import static org.junit.Assert.fail;
+import org.junit.Test;
+
+public class QueryTest {
+  @Test
+  public void testParsingQuery_WithNullQueryType_ThrowsNullPointerException() {
+    try {
+      Query query = new Query(/*type = */ null, "com.google.cats.Cat");
+      fail();
+    } catch (NullPointerException e) {
```
kwchan1995: [`assertThrows`](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertThrows-java.lang.Class-org.junit.jupiter.api.function.Executable-) would be more idiomatic :smiley:

project/tests/com/google/daggerquery/executor/services/DaggerQueryControllerTest.java
```diff
@@ -0,0 +1,129 @@
...
+import static org.junit.Assert.fail;
+
+import com.google.daggerquery.protobuf.autogen.BindingGraphProto;
+import com.google.daggerquery.protobuf.autogen.DependencyProto;
+import org.junit.Test;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.PrintStream;
+
+public class DaggerQueryControllerTest {
```
kwchan1995: It appears to me this test largely overlaps with `QueryExecutorTest`, and tests the exact output in addition. How about just keeping one of them to reduce the maintenance overhead?

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -0,0 +1,67 @@
...
+limitations under the License.
+ */
+
+package com.google.daggerquery.executor.models;
+
+/**
+ * A class which represents one query.
+ *
+ * <p>Each query has a {@code type} which itself an instance of {@link QueryType}.
+ * There cannot be two queries with the same type.
```
kwchan1995: I'm confused by this sentence :confused: 

It seems to suggest I cannot ask `deps Cat` followed by `deps Dogs`. Maybe remove?

project/src/com/google/daggerquery/executor/services/DaggerQueryController.java
margaritiko: Just to clarify: for example, for this particular case with `ArrayIndexOutOfBoundsException` I can at first check the size of passed array with arguments and then throw exception with detailed message about error. I fixed it this way ✅

project/src/com/google/daggerquery/executor/services/QueryExecutor.java
margaritiko: Actually I have just checked and my implementation of saving graph already associates a leaf with an empty array, so I fixed logic of making a fake graph in tests 👌

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
margaritiko: Yes, it makes sense! Thanks. Now all dependencies are written on separate lines without additional comments.

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: I mean, `QueryType` can be the id of each query, but yes, I wrote it not clear and this sentence is quite ambiguous. Anyway, I am saying about this fact in javadoc for `QueryType` class. Removed ✅

project/tests/com/google/daggerquery/executor/services/DaggerQueryControllerTest.java
margaritiko: Yes, there are a big intersection in cases which are tested by these two classes. The main purpose of the `QueryExecutorTest` implementation was to test the actual behavior of a program when a user uses it, but it was mostly testing an output, which isn't good. I removed it.

project/tests/com/google/daggerquery/executor/models/QueryTest.java
margaritiko: The reason I chose this approach with try/catch is because `assertThrows` was supported only in JUnit 5.

As described in [blog](https://flexport.engineering/connecting-bazel-and-junit5-by-transforming-arguments-46440c6ea068), a solution for supporting JUnit 5 is following: adding `junit5.bzl` file from [here](https://github.com/junit-team/junit5-samples/tree/main/junit5-jupiter-starter-bazel) and using `java_junit5_test` instead of `java_test`.

I thought that it can make migration of project from open-source a little harder and will increase the total number of lines with this file `junit5.bzl` (which isn't a project enhancement).

How do you think?

project/src/com/google/daggerquery/executor/services/InternalSourcesLoader.java
margaritiko: Done 😃

project/tests/com/google/daggerquery/executor/models/QueryTest.java
kwchan1995: Ah sorry, I didn't realise this is a new feature, which seems to have been added since JUnit 4.13 (https://github.com/junit-team/junit4/blob/HEAD/doc/ReleaseNotes4.13.md).

Would it be straightforward to bump the JUnit version from 4.11 to 4.13 here:

https://github.com/googleinterns/dagger-query/blob/6f9f1b834ccf8876bbc4198d9132a3e39c06e21a/project/workspace_defs.bzl#L91

If not, how about using the `@Test(expected = ...Exception.class)` annotation, like this [StackOverflow post](https://stackoverflow.com/a/156528)? :smiley:

kwchan1995 (approved):
Awesome milestone!

project/src/com/google/daggerquery/executor/BUILD
```diff
@@ -0,0 +1,29 @@
...
+java_binary(
+    name = "query_executor",
+    srcs = ["QueryExecutorMain.java"],
+    main_class = "com.google.daggerquery.executor.QueryExecutorMain",
+    deps = [
+        "//src/com/google/daggerquery/executor/models:query_executor_models",
+        "//src/com/google/daggerquery/executor/services:query_executor_services",
+        "//src/com/google/daggerquery/protobuf:dagger_query_protobuf"
+    ],
+    resources = ["//src/com/google/daggerquery:binding_graph_data"],
```
Ratchette, line 28: I can't seem to find this build target. 
Is this something leftover from a previous commit?

project/src/com/google/daggerquery/executor/QueryExecutorMain.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.google.daggerquery.executor.services.InternalSourcesLoader;
+import com.google.daggerquery.executor.services.DaggerQueryController;
+
+public class QueryExecutorMain {
+  public static void main(String[] args) {
+    new DaggerQueryController().execute(
+        /*arguments = */    args,
+        /*printStream = */  System.out,
+        /*sourcesLoader =*/ new InternalSourcesLoader()
```
Ratchette: Nit: would you mind making this consistent (eg. adding a space to the sourcesLoader comment or removing a space from the two others?

project/src/com/google/daggerquery/executor/BUILD
Ratchette: I can't seem to find this build target.
Is this something leftover from a previous commit?

project/src/com/google/daggerquery/executor/services/InternalSourcesLoader.java
```diff
@@ -0,0 +1,37 @@
...
+package com.google.daggerquery.executor.services;
+
+import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+
+public class InternalSourcesLoader implements SourcesLoader {
+  private static final String PATH_TO_BINDING_GRAPH = "/com/google/daggerquery/binding_graph_data.textproto";
```
Ratchette: Is this path relative to bazel-out?
I'm having trouble tracking where this file is being created. Will it be generated by the dagger_query_textproto build rule? 
https://github.com/googleinterns/dagger-query/blob/131eaa5c23d4f2654205ab3938ed7075b0c0de21/project/src/com/google/daggerquery/dagger_query_textproto.bzl#L15

project/src/com/google/daggerquery/executor/services/InternalSourcesLoader.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public BindingGraph loadBindingGraph() throws IOException {
+    InputStream inputStream = InternalSourcesLoader.class.getResourceAsStream(PATH_TO_BINDING_GRAPH);
+
+    if (inputStream == null) {
+      throw new FileNotFoundException(String.format("File %s is missed.", PATH_TO_BINDING_GRAPH));
```
Ratchette: Grammatical nit:
A more commonly used error phrasing would be:
"File %s is **missing**."

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.io.PrintStream;
+import java.util.List;
+
+public class QueryOutputWriter {
+  private PrintStream printStream;
+
+  /**
+   * Creates an instance of {@link QueryOutputWriter} and
+   * associates it with given {@link PrintStream} instance,
```
Ratchette: grammatical nit:
... associates it with **the** given ...

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
```diff
@@ -0,0 +1,43 @@
...
+   * where all output will be written.
+   */
+  public QueryOutputWriter(PrintStream printStream) {
+    this.printStream = printStream;
+  }
+
+  /**
+   * Writes an output of execution {@link com.google.daggerquery.executor.models.QueryType}.DEPS query.
```
Ratchette: Will there be different QueryOutputWriters for different queries?

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * <p>An output is an array with all dependencies presented as Strings,
```
Ratchette: Grammatical nit: 
**The** output ....

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
```diff
@@ -0,0 +1,43 @@
...
+   * so method simply writes them one by one.
```
Ratchette: Grammatical nit: 
so **this** method ....

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
```diff
@@ -0,0 +1,43 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+ */
+
+package com.google.daggerquery.executor.services;
+
+import java.io.PrintStream;
+import java.util.List;
+
+public class QueryOutputWriter {
```
Ratchette: At the moment, this class looks like a wrapper for a printStream.
What is the advantage of using this class over using the printStream directly?

project/src/com/google/daggerquery/executor/services/QueryExecutor.java
```diff
@@ -0,0 +1,50 @@
...
+import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph;
+import com.google.daggerquery.protobuf.autogen.DependencyProto;
+
+import java.io.PrintStream;
+import java.util.ArrayList;
+import java.util.List;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Special class for executing queries by traversing a {@link BindingGraph} instance.
```
Ratchette: Optional Nit:
A class for executing queries on a {@link BindingGraph}

project/src/com/google/daggerquery/executor/services/QueryExecutor.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * <p>Responsible for executing query, which includes finding the result
+ * and writing it for user in specified {@link PrintStream} instance.
```
Ratchette: It looks like now this class doesn't interact with the PrintStream directly, and instead returns a List of Strings to the calling method.

Please update or delete this comment.

project/src/com/google/daggerquery/executor/models/QueryType.java
```diff
@@ -0,0 +1,60 @@
...
+
+package com.google.daggerquery.executor.models;
+
+/**
+ * An enum which represents type of query, which can be executed via traversing binding graph.
+ *
+ * <p>Each query identified by its name and number of parameters it takes.
+ * This data associated with {@link QueryType} instance.
+ */
+public enum QueryType {
```
Ratchette: Could this class be represented by a map of name -> # deps instead?

project/tests/com/google/daggerquery/executor/models/QueryTest.java
margaritiko: For some reason changing version didn't help, but using this annotation with `expected` keyword is awesome, thanks! I didn't know about it before, tests look really better now 😊

project/src/com/google/daggerquery/executor/QueryExecutorMain.java
margaritiko: Sure! Thanks 😊

project/src/com/google/daggerquery/executor/services/InternalSourcesLoader.java
margaritiko: Yes, exactly. This file is created by `dagger_query_textproto`. I have removed resources attribute now and will add [in different PR ](https://github.com/googleinterns/dagger-query/projects/1#card-43589741) all instructions.

project/src/com/google/daggerquery/executor/BUILD
margaritiko: Removed and will be described in [a next PR](https://github.com/googleinterns/dagger-query/projects/1#card-43589741).

project/src/com/google/daggerquery/executor/services/InternalSourcesLoader.java
margaritiko: Thanks!

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
margaritiko: ✅

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
margaritiko: ✅

project/src/com/google/daggerquery/executor/services/QueryExecutor.java
margaritiko: ✅

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
margaritiko: ✅

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
margaritiko: I removed this class and put all logic with writing an output in a main method. Now it doesn't depend on query type.

project/src/com/google/daggerquery/executor/services/QueryOutputWriter.java
margaritiko: Yes, I agree with you - it's too complicated for this moment. I removed this class and now output doesn't depend on query type.

project/src/com/google/daggerquery/executor/services/QueryExecutor.java
margaritiko: Class was removed ✅

project/src/com/google/daggerquery/executor/models/QueryType.java
margaritiko: Yes, good idea! I made this map and put it in `Query` class.

project/src/com/google/daggerquery/executor/QueryExecutor.java
```diff
@@ -0,0 +1,54 @@
...
+   *
+   * <p>The first argument must always define the name of a query.
+   * Next arguments belong to a query as its parameters.
+   */
+  public static void main(String[] args) {
+    PrintStream printStream = System.out;
+
+    try {
+      if (args.length == 0) {
+        throw new IllegalArgumentException("You might forgot to specify the request and its parameters.");
```
Ratchette: Nit:
"You did not specify ...."

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -0,0 +1,98 @@
...
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * A class which represents a query.
+ *
+ * <p>Each query has a String {@code name} value.
+ *
+ * <p>Also each query contains parameters which were passed by user.
+ * The number of parameters completely defined by query's name.
```
The number of parameters **is** completely defined by **the** query's name.

project/tests/com/google/daggerquery/plugin/GraphConverterTest.java
```diff
@@ -67,47 +67,27 @@ public void testMakingGraph_WhenContainsParallelEdges() {
     assertEquals(0, adjacencyList.get("B").getDependencyCount());
   }
 
-  @Test
+  @Test(expected = IllegalArgumentException.class)
```
Ratchette, line 70: This is a really nice improvement 🏄‍♀️

project/tests/com/google/daggerquery/executor/models/QueryTest.java
```diff
@@ -0,0 +1,136 @@
...
+  @Test(expected = NullPointerException.class)
+  public void testExecutingDepsQuery_WithNullBindingGraph_ThrowsNullPointerException() {
+    String[] parameters = {"com.google.Cat"};
+    Query query = new Query("deps", parameters);
+
+    query.execute(null);
+  }
+
+  /*
+   * Makes a simple non-cycled binding graph with the following structure:
```
Ratchette: Nit:
Makes a simple **acyclic** binding graph ....

project/tests/com/google/daggerquery/executor/models/QueryTest.java
```diff
@@ -0,0 +1,136 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+import com.google.daggerquery.protobuf.autogen.BindingGraphProto;
+import com.google.daggerquery.protobuf.autogen.DependencyProto;
+import org.junit.Test;
+
+import java.util.List;
+
+public class QueryTest {
```
Ratchette, line 29: You did a wonderfully thorough job with these tests.

Ratchette (approved):
LGTM!

project/tests/com/google/daggerquery/executor/models/QueryTest.java
margaritiko: Thanks ✔️

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: ✔️

project/src/com/google/daggerquery/executor/QueryExecutor.java
margaritiko: ✔️

PR #13, opened 2020-08-11
Automagically generate protobufs using Bazel's java_proto_library rule.

Ratchette:


Ratchette:
There are 3 main changes to this CL:
1) Update the version of proto_library & java_proto_library we're using (in workspace_defs.bzl)
2) Add additional protobuf deps (in the WORKSPACE file)
3) Split the proto generation into two separate rules (BUILD files)

The remaining files are minor cleanup changes related to this CL.

project/src/com/google/daggerquery/protobuf/BUILD
```diff
@@ -12,12 +12,28 @@
...
 
-java_library(
-    name = "dagger_query_protobuf",
-    srcs = glob(["autogen/*.java"]),
-    deps = ["//third_party/java/protobuf:protobuf"],
+java_proto_library(
+    name = "binding_graph_java_proto",
+    deps = [
+        ":binding_graph_proto",
+        ":dependency_proto"
```
kwchan1995: I suspect `:dependency_proto` may not be required. Would you mind trying to remove and test if it continues to build?

(I mean just for `java_proto_library`)

Ratchette: You are totally right. I tested it out, and this dependency is only needed for the proto_library rule.

margaritiko:
Maybe now we can also remove [autogen files](https://github.com/googleinterns/dagger-query/tree/protos/project/src/com/google/daggerquery/protobuf/autogen) from src/com/google/daggerquery/protobuf?

kwchan1995 (approved):

Ratchette:
> Maybe now we can also remove [autogen files](https://github.com/googleinterns/dagger-query/tree/protos/project/src/com/google/daggerquery/protobuf/autogen) from src/com/google/daggerquery/protobuf?

Good idea - I've added a commit to this PR that deletes those files.

margaritiko (approved):
Great ✨

PR #15, opened 2020-08-14
Supported ALLPATHS query 🚀

margaritiko:
Support `allpaths` query and cover it with tests

#### Main changes in this PR:
1. Added two constants for containing queries names in `Query` class to avoid using simple strings several times in code.
2. Extended `public execute(...)` method in `Query` class with logic for executing **allpaths** query.
3.  Implemented helper method `findAllPaths(...)` in `Query` class which recursively traverses a graph and constructs all possible pathes.
4. Added tests into `QueryTest` class.

[Card №1](https://github.com/googleinterns/dagger-query/projects/1#card-42487520)
[Card №2](https://github.com/googleinterns/dagger-query/projects/1#card-42487638)

### DIFF
--- a/project/src/com/google/daggerquery/executor/models/Query.java
+++ b/project/src/com/google/daggerquery/executor/models/Query.java
@@ -17,12 +17,14 @@ limitations under the License.
 package com.google.daggerquery.executor.models;
 
 import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph;
-import com.google.daggerquery.protobuf.autogen.DependencyProto;
+import com.google.daggerquery.protobuf.autogen.DependencyProto.Dependency;
 import sun.reflect.generics.reflectiveObjects.NotImplementedException;
-
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import static java.util.stream.Collectors.toList;
 
@@ -35,13 +37,18 @@ import static java.util.stream.Collectors.toList;
  * The number of parameters is completely defined by the query's name.
  */
 public class Query {
+
+  private final static String DEPS_QUERY_NAME = "deps";
+  private final static String ALLPATHS_QUERY_NAME = "allpaths";
+
   /**
    * The key is the name of supported query and the value is the number of parameters,
    * required by query with such name.
    */
   private final static Map<String, Integer> supportedQueries = new HashMap<String, Integer>() {
     {
-      put("deps", 1);
+      put(DEPS_QUERY_NAME, 1);
+      put(ALLPATHS_QUERY_NAME, 2);
     }
   };
 
@@ -83,16 +90,52 @@ public class Query {
    * <p>For `deps` queries each string represent exactly one dependency.
    */
   public List<String> execute(BindingGraph bindingGraph) {
-    if (name.equals("deps")) {
-      String sourceNode = parameters[0];
-      if (!bindingGraph.getAdjacencyListMap().containsKey(sourceNode)) {
-        throw new IllegalArgumentException("Specified source node doesn't exist.");
-      }
+    String source = parameters[0];
 
-      return bindingGraph.getAdjacencyListMap().get(sourceNode).getDependencyList()
-          .stream().map(DependencyProto.Dependency::getTarget).sorted().collect(toList());
+    if (!bindingGraph.getAdjacencyListMap().containsKey(source)) {
+      throw new IllegalArgumentException("Specified source node doesn't exist.");
+    }
+
+    switch (name) {
+      case DEPS_QUERY_NAME:
+        return bindingGraph.getAdjacencyListMap().get(source).getDependencyList()
+            .stream().map(Dependency::getTarget).sorted().collect(toList());
+
+      case ALLPATHS_QUERY_NAME:
+        String target = parameters[1];
+        Set<String> visitedNodes = new HashSet<>();
+        List<List<String>> result = new ArrayList<>();
+        List<String> path = new ArrayList<>();
+
+        findAllPaths(source, target, path, bindingGraph, visitedNodes, result);
+        return result.stream().map(list -> String.join(" -> ", list)).collect(toList());
     }
 
     throw new NotImplementedException();
+  }
+
+  private void findAllPaths(String source, String target, List<String> path, BindingGraph bindingGraph,
+                            Set<String> visitedNodes, List<List<String>> allPaths) {
+    path.add(source);
+
+    // If we've already found a path from `source` to `target`, we can stop and not go deeper.
+    if (source.equals(target)) {
+      allPaths.add(new ArrayList<>(path));
+      path.remove(source);
+      return;
+    }
+
+    visitedNodes.add(source);
+    System.out.println(String.format("Visiting %s", source));
+    for (Dependency nextNode: bindingGraph.getAdjacencyListMap().get(source).getDependencyList()) {
+      if (visitedNodes.contains(nextNode.getTarget())) {
+        continue;
+      }
+
+      findAllPaths(nextNode.getTarget(), target, path, bindingGraph, visitedNodes, allPaths);
+    }
+
+    visitedNodes.remove(source);
+    path.remove(source);
   }
 }
--- a/project/src/com/google/daggerquery/executor/models/Query.java
+++ b/project/src/com/google/daggerquery/executor/models/Query.java
@@ -132,7 +132,7 @@ public class Query {
    *
    * <p>Uses delimiter '->' for constructing the string representation of all data.
    */
-  private class Path<NodeT> {
+  private static class Path<NodeT> {
     private Deque<NodeT> nodesDeque = new ArrayDeque<>();
 
     Path() {
--- a/project/src/com/google/daggerquery/executor/models/Query.java
+++ b/project/src/com/google/daggerquery/executor/models/Query.java
@@ -95,18 +95,23 @@ public class Query {
    * </ul>
    */
   public List<String> execute(BindingGraph bindingGraph) {
-    String source = parameters[0];
-
-    if (!bindingGraph.getAdjacencyListMap().containsKey(source)) {
-      throw new IllegalArgumentException("Specified source node doesn't exist.");
-    }
-
     switch (name) {
       case DEPS_QUERY_NAME:
-        return bindingGraph.getAdjacencyListMap().get(source).getDependencyList()
-            .stream().map(Dependency::getTarget).sorted().collect(toList());
+        String sourceNode = parameters[0];
 
+        if (!bindingGraph.getAdjacencyListMap().containsKey(sourceNode)) {
+          throw new IllegalArgumentException("Specified source node doesn't exist.");
+        }
+
+        return bindingGraph.getAdjacencyListMap().get(sourceNode).getDependencyList()
+            .stream().map(Dependency::getTarget).sorted().collect(toList());
       case ALLPATHS_QUERY_NAME:
+        String source = parameters[0];
+
+        if (!bindingGraph.getAdjacencyListMap().containsKey(source)) {
+          throw new IllegalArgumentException("Specified source node doesn't exist.");
+        }
+
         String target = parameters[1];
         Set<String> visitedNodes = new HashSet<>();
         List<List<String>> result = new ArrayList<>();
--- a/project/src/com/google/daggerquery/executor/models/Query.java
+++ b/project/src/com/google/daggerquery/executor/models/Query.java
@@ -86,8 +86,14 @@ public class Query {
   /**
    * Executes query on a {@link BindingGraph}.
    *
-   * <p>For all queries return a list with string. Strings content depends on a query name.
-   * <p>For `deps` queries each string represent exactly one dependency.
+   * <p>For all queries return a list with strings. Strings content depends on a query name.
+   *
+   * <ul>
+   * <li>For `deps` query each string represent exactly one dependency.
+   * <li>For `allpaths` query each string contains a path between {@code source} and {@code target} nodes.
+   * The connection between nodes is shown with construction '->'. For example, one of the possible paths
+   * may look like this: "com.google.First -> com.google.Second -> com.google.Third".
+   * </ul>
    */
   public List<String> execute(BindingGraph bindingGraph) {
     String source = parameters[0];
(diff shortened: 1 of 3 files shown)


project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -79,20 +86,71 @@ public Query(String typeName, String... parameters) {
...
+        return bindingGraph.getAdjacencyListMap().get(source).getDependencyList()
+            .stream().map(Dependency::getTarget).sorted().collect(toList());
+
+      case ALLPATHS_QUERY_NAME:
+        String target = parameters[1];
+        Set<String> visitedNodes = new HashSet<>();
+        List<List<String>> result = new ArrayList<>();
+        List<String> path = new ArrayList<>();
+
+        findAllPaths(source, target, path, bindingGraph, visitedNodes, result);
```
margaritiko, line 121: We care about time complexity more, so it’s a reason for using the `visitedNodes` set: with it we can check a node with **O(1)**, but if we will search in a `path` for a node it will take us **O(n)**, where `n` is a length of a path, which might be big

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -79,20 +86,71 @@ public Query(String typeName, String... parameters) {
   /**
    * Executes query on a {@link BindingGraph}.
    *
-   * <p>For all queries return a list with string. Strings content depends on a query name.
-   * <p>For `deps` queries each string represent exactly one dependency.
+   * <p>For all queries return a list with strings. Strings content depends on a query name.
```
**The content of the strings** depends on a query name

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -79,20 +86,71 @@ public Query(String typeName, String... parameters) {
...
+   *
+   * <ul>
+   * <li>For `deps` query each string represent exactly one dependency.
```
For \`deps\` query each string **represents** exactly one dependency.

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -79,20 +86,71 @@ public Query(String typeName, String... parameters) {
...
+   * <li>For `allpaths` query each string contains a path between {@code source} and {@code target} nodes.
+   * The connection between nodes is shown with construction '->'. For example, one of the possible paths
```
with **the** construction

project/tests/com/google/daggerquery/executor/models/QueryTest.java
```diff
@@ -125,12 +218,57 @@ public void testExecutingDepsQuery_WithNullBindingGraph_ThrowsNullPointerExcepti
...
+        .build();
+  }
+
+  /*
+   * Makes an acyclic binding graph with multiple paths between nodes with the following structure:
+   *
+   * com.google.Component --> com.google.CatsFactory --> com.google.Cat --> com.google.Details
+   * com.google.Component --> com.google.Helper --> com.google.CatsFactory --> com.google.Cat --> com.google.Details
+   * com.google.Component --> com.google.Cat --> com.google.Details
+   * com.google.Component --> com.google.Details
```
Ratchette, line 242: This is great. Having the graph described in the comments makes this code much easier to understand.

project/tests/com/google/daggerquery/executor/models/QueryTest.java
```diff
@@ -28,6 +28,8 @@
 
 public class QueryTest {
 
+  // Tests for `DEPS` query
+
```
Ratchette: Would you mind moving this to be one test lower - above the testParsingDepsQuery_WithTwoStringParameters_ThrowsIllegalArgumentException test?

This test (where the query name is not provided) looks like it can be considered common testing infrastructure across all query types.

project/tests/com/google/daggerquery/executor/models/QueryTest.java
```diff
@@ -107,6 +109,97 @@ public void testExecutingDepsQuery_WithNullBindingGraph_ThrowsNullPointerExcepti
...
+    Query query = new Query("allpaths", parameters);
+
+    List<String> queryExecutionResult = query.execute(makeBindingGraph_WithMultiplePathsBetweenTwoNodes());
+
+    String[] expectedOutput = {
+        "com.google.Component -> com.google.CatsFactory -> com.google.Cat -> com.google.Details",
+        "com.google.Component -> com.google.Helper -> com.google.CatsFactory -> com.google.Cat -> com.google.Details",
+        "com.google.Component -> com.google.Cat -> com.google.Details",
+        "com.google.Component -> com.google.Details"
+    };
```
Ratchette: Can we guarantee that the ordering of these Strings be deterministic? 
(AKA that they will be in the same order each time)?

I think this can depend on a number of things:
1) How dagger constructs it's binding graph.
2) How dagger plugins traverse the binding graph.
3) How protobufs store data. 
4) If the algorithm for the query executor is deterministic.

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -35,13 +37,18 @@
...
    * The key is the name of supported query and the value is the number of parameters,
    * required by query with such name.
    */
   private final static Map<String, Integer> supportedQueries = new HashMap<String, Integer>() {
     {
-      put("deps", 1);
+      put(DEPS_QUERY_NAME, 1);
+      put(ALLPATHS_QUERY_NAME, 2);
     }
   };
```
kwchan1995: Consider using Guava's [`ImmutableMap`](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ImmutableMap.html), which supports value initialisation with the `.of()` operator or `.builder()`.

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -79,20 +86,71 @@ public Query(String typeName, String... parameters) {
...
+   * may look like this: "com.google.First -> com.google.Second -> com.google.Third".
+   * </ul>
    */
   public List<String> execute(BindingGraph bindingGraph) {
-    if (name.equals("deps")) {
-      String sourceNode = parameters[0];
-      if (!bindingGraph.getAdjacencyListMap().containsKey(sourceNode)) {
-        throw new IllegalArgumentException("Specified source node doesn't exist.");
-      }
+    String source = parameters[0];
```
kwchan1995: I'd prefer inlining this in the individual `case` clauses below for extensibility. While this is true today, future queries that we support may not have the first parameter as the source binding.

I understand this will introduce a bit of duplication, but I think this is a fair tradeoff against maintainability.

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -79,20 +86,71 @@ public Query(String typeName, String... parameters) {
...
 
     throw new NotImplementedException();
   }
+
+  /**
+   * Traverses a {@link BindingGraph} starting from {@code source} node.
+   *
+   * <p>Puts all processed nodes in a {@code visitedNodes} set to avoid loops. Constructs a {@code path} which is
+   * an instance of {@link List<String>}. At each recursive level this variable contains a correct path in a graph from
+   * some root node which was passed in the first call and {@code target} node which is the same for all calls.
```
kwchan1995, line 164: Do you mean the *previous* call rather than the *first* call? (which I interpret it to be the `source` node tied to the query) :stuck_out_tongue:

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -79,20 +86,71 @@ public Query(String typeName, String... parameters) {
...
+   *
+   * <p>As the result fills provided {@link List<List<String>>} {@code allPaths} with all possible paths between
+   * root and target nodes.
+   */
+  private void findAllPaths(String source, String target, List<String> path, BindingGraph bindingGraph,
+                            Set<String> visitedNodes, List<List<String>> allPaths) {
```
kwchan1995: I'd recommend changing the type signature to Guava's `ImmutableList.Builder<List<String>>`, which seems clearer in intent.

Or consider a layer of abstraction like `ImmutableList.Builder<Path>`, where `Path` is a helper class that wraps your list representation. It also makes it possible to add more properties to `Path` later or refactor its internal implementation should you need to.

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -79,20 +86,71 @@ public Query(String typeName, String... parameters) {
...
+    for (Dependency nextNode: bindingGraph.getAdjacencyListMap().get(source).getDependencyList()) {
+      if (visitedNodes.contains(nextNode.getTarget())) {
+        continue;
+      }
+
+      findAllPaths(nextNode.getTarget(), target, path, bindingGraph, visitedNodes, allPaths);
+    }
+
+    visitedNodes.remove(source);
+    path.remove(source);
```
kwchan1995: The `.remove(Object)` operation on a `List` costs `O(n)`. If you only operate on the two ends, how about using an alternative data structure like an `ArrayDeque`?

project/tests/com/google/daggerquery/executor/models/QueryTest.java
```diff
@@ -125,12 +218,57 @@ public void testExecutingDepsQuery_WithNullBindingGraph_ThrowsNullPointerExcepti
...
+        .build();
+    BindingGraphProto.BindingGraph.ListWithDependencies catNodeDeps = BindingGraphProto.BindingGraph.ListWithDependencies.newBuilder()
+        .addDependency(detailsNode)
+        .build();
+    BindingGraphProto.BindingGraph.ListWithDependencies detailsNodeDeps = BindingGraphProto.BindingGraph.ListWithDependencies.newBuilder()
+        .build();
 
     return BindingGraphProto.BindingGraph.newBuilder()
         .putAdjacencyList("com.google.Component", componentNodeDeps)
         .putAdjacencyList("com.google.CatsFactory", factoryNodeDeps)
```
kwchan1995: nit: Please rename this to `catsFactoryNodeDeps` for consistency.

project/tests/com/google/daggerquery/executor/models/QueryTest.java
margaritiko: Done, I've moved test outside these `deps` tests  section👌

project/tests/com/google/daggerquery/executor/models/QueryTest.java
margaritiko: In this test, we can guarantee that the order is the same each launch. That's because the graph is created exactly in this class below and we skip your first three points. The algorithm for the query executor will always return the same answer for the same input.

But for a general case, it's harder to predict ordering because of the reasons you mentioned.

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: Looks nice, thank you! The `.of()` operator allows adding only up to 5 key-value pairs so I decided to use `.builder()` instead to avoid changing it in future when more queries will be supported.

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: Sounds reasonable, I will move it, thank you 🌟

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: For a _previous_ node, it's also true, but in general, it's true also for the first passed node, because it's exactly what we need - to get all paths between the first passed as `source` node and a `target`.

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: Oh, yes, exactly! As it behaves as a stack `ArrayDeque` is pretty good, I've added it.

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: I've added a generic class `Path` and also overrided `toString()` in it to move logic of constructing string representation of a path into path instance itself ✅

margaritiko (commented):
@Ratchette @kwchan1995 
I made changes, and it is now up to date ☀️

project/src/com/google/daggerquery/executor/models/Query.java
kwchan1995: This may be more of a hack (which I have used myself), is to use Guava's [`ImmutableSet`](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/ImmutableSet.html). Guava's [`ImmutableCollection`](https://github.com/google/guava/wiki/ImmutableCollectionsExplained#how) classes guarantee the iteration order to be the same as construction order. You may be able to cheekily use this for the `Path` implementation, if your `Path` representation never contains duplicate elements (cycles).

project/src/com/google/daggerquery/executor/models/Query.java
kwchan1995: Sorry you're right, I misinterpreted the code.

project/src/com/google/daggerquery/executor/models/Query.java
kwchan1995: Sorry I forgot that I proposed using `ArrayDeque` earlier. Yeah I'm not sure if there 's a data structure which implements both `Set` and `Stack`. Please ignore.

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -79,20 +86,107 @@ public Query(String typeName, String... parameters) {
...
+  }
+
+  /**
+   * Represents a path between nodes of type {@code NodeT}.
+   *
+   * <p>Supports operations of adding a new node to the back and removing the last node.
+   *
+   * <p>Uses delimiter '->' for constructing the string representation of all data.
+   */
+  private class Path<NodeT> {
```
kwchan1995: Please make this a `static` inner class, because it does not require access to state variables in the outer class.

kwchan1995: Alternatively it may make sense to define `Path` in a separate file from `Query`, if they are two different concepts.

kwchan1995 (approved):

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: Thanks! I would prefer to save it as an inner class because it is closely related to the query's output (i.e. it's tested as a part of `Query` class and used only in it).

project/tests/com/google/daggerquery/executor/models/QueryTest.java
margaritiko: I've [changed the test](https://github.com/googleinterns/dagger-query/pull/15/commits/85d91497ef42eb14ebf1fdc7fd1faa79c90d0320) to ignore an ordering because it doesn't matter in which order paths are listed so this one-to-one comparison is not necessary.

Ratchette (approved):
LGTM 🌯

PR #17, opened 2020-08-18
Serialize all binding graphs into different files and pack them into one .zip archive 👩‍💻

margaritiko:
Saves all .textproto files into one .zip file

 ### Problem:
 In real applications, there can be multiple binding graphs, because nothing prevents the user from having different components. In this situation method `visitGraph(BindingGraph bindingGraph, DiagnosticReporter diagnosticReporter)` in `QueryPlugin` is calling several times - one call for each graph exactly. With the current solution, only the last graph is saved because it simply fills the resource file only once.

 ### Solution:
 The idea is to save all binding graphs in the .zip file and then extract .textproto files from it in client apps. 

  **Main changes:** 
 * in implementation of `dagger_query_textproto.bzl` changed command - now it unzips all files with .textproto extension and puts them into new zip package;
 * in `QueryPlugin` saves binding graphs into different files which name is constructed via component's name.

 ### Alternatives:
 I thought about making a new proto-model - a wrapper for a list of `BindingGraph` instances. But it's impossible to change the content of resource file which was created on the previous step: [Filer](https://docs.oracle.com/javase/7/docs/api/javax/annotation/processing/Filer.html) class has a method `getResource(...)` **but it allows to open a file for reading, not for writing 😢** 

[Card](https://github.com/googleinterns/dagger-query/projects/1#card-43510033)

### DIFF
--- a/project/src/com/google/daggerquery/dagger_query_textproto.bzl
+++ b/project/src/com/google/daggerquery/dagger_query_textproto.bzl
@@ -25,7 +25,7 @@ def _dagger_query_textproto_impl(ctx):
     ctx.actions.run_shell(
         inputs = src_jars,
         outputs = [ctx.outputs.out],
-        command = "unzip -p {src_jar} binding_graph.textproto > {out}".format(
+        command = "touch {out}; unzip {src_jar}; for FILE in *.textproto; do cat $FILE >> {out}; done".format(
             src_jar = src_jars[0].path,
             out = ctx.outputs.out.path,
         ),
--- a/project/src/com/google/daggerquery/dagger_query_textproto.bzl
+++ b/project/src/com/google/daggerquery/dagger_query_textproto.bzl
@@ -25,7 +25,7 @@ def _dagger_query_textproto_impl(ctx):
     ctx.actions.run_shell(
         inputs = src_jars,
         outputs = [ctx.outputs.out],
-        command = "unzip {src_jar} '*.textproto' -d .; zip -R {out} '*.textproto'".format(
+        command = "unzip {src_jar} '*.textproto' -d .; zip {out} '*.textproto'".format(
             src_jar = src_jars[0].path,
             out = ctx.outputs.out.path,
         ),
--- a/project/src/com/google/daggerquery/dagger_query_textproto.bzl
+++ b/project/src/com/google/daggerquery/dagger_query_textproto.bzl
@@ -25,7 +25,7 @@ def _dagger_query_textproto_impl(ctx):
     ctx.actions.run_shell(
         inputs = src_jars,
         outputs = [ctx.outputs.out],
-        command = "unzip {src_jar} '*.textproto' -d .; zip {out} '*.textproto'".format(
+        command = "unzip {src_jar} '*_graph.textproto' -d .; zip {out} '*_graph.textproto'".format(
             src_jar = src_jars[0].path,
             out = ctx.outputs.out.path,
         ),
--- a/project/src/com/google/daggerquery/dagger_query_textproto.bzl
+++ b/project/src/com/google/daggerquery/dagger_query_textproto.bzl
@@ -25,7 +25,7 @@ def _dagger_query_textproto_impl(ctx):
     ctx.actions.run_shell(
         inputs = src_jars,
         outputs = [ctx.outputs.out],
-        command = "touch {out}; unzip {src_jar}; for FILE in *.textproto; do cat $FILE >> {out}; done".format(
+        command = "unzip {src_jar} '*.textproto' -d .; zip -R {out} '*.textproto'".format(
             src_jar = src_jars[0].path,
             out = ctx.outputs.out.path,
         ),
@@ -39,7 +39,7 @@ _dagger_query_textproto = rule(
         ),
     },
     outputs = {
-        "out": "%{name}.textproto",
+        "out": "%{name}.zip",
     },
     implementation = _dagger_query_textproto_impl,
 )
--- a/project/src/com/google/daggerquery/executor/services/SourcesLoader.java
+++ b/project/src/com/google/daggerquery/executor/services/SourcesLoader.java
@@ -20,6 +20,7 @@ import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.ByteBuffer;
 
 /**
  * A class which loads a binding graph saved with Dagger SPI plugin.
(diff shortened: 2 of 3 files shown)


project/src/com/google/daggerquery/executor/services/SourcesLoader.java
```diff
@@ -28,12 +29,13 @@
...
   public BindingGraph loadBindingGraph() throws IOException {
-    InputStream inputStream = SourcesLoader.class.getResourceAsStream(PATH_TO_BINDING_GRAPH);
+    try (InputStream inputStream = SourcesLoader.class.getResourceAsStream(PATH_TO_BINDING_GRAPH))  {
+      if (inputStream == null) {
+        throw new FileNotFoundException(String.format("File %s is missing.", PATH_TO_BINDING_GRAPH));
+      }
 
-    if (inputStream == null) {
-      throw new FileNotFoundException(String.format("File %s is missing.", PATH_TO_BINDING_GRAPH));
+      int sizeOfBindingGraph = ByteBuffer.wrap(inputStream.readNBytes(4)).getInt();
```
kwchan1995: Where does the constant `4` come from?

margaritiko: It's the number of bytes required to save an integer (i.e. 32 / 8). 

Actually, I'm not sure that 2 147 483 647 (max int value) is enough to store the huge graphs in AGSA and I don't know how to check it now.

kwchan1995:
> It looks attractive to save binding graphs into different files, put them into one specific directory and then use all its content. However from this thread I understood that it's impossible to say that output of the rule is a directory with all its content. Maybe something changed now but I didn't find info about this feature in docs.

Maybe I'm missing something - I thought everything you write using the filer will get packaged into the same jar, and therefore you should be able to retrieve them from within the [`shell command`](https://github.com/googleinterns/dagger-query/pull/17/files#diff-984f141cfa58d26ca608b4c2b15d488bR28)?

margaritiko:
> Maybe I'm missing something - I thought everything you write using the filer will get packaged into the same jar, and therefore you should be able to retrieve them from within the [`shell command`](https://github.com/googleinterns/dagger-query/pull/17/files#diff-984f141cfa58d26ca608b4c2b15d488bR28)?

I think that the problem is here: [in this line](https://github.com/googleinterns/dagger-query/blob/d7a7815e1da67d12217bf6f6a33d0c62ed88701a/project/src/com/google/daggerquery/dagger_query_textproto.bzl#L42) we explicitly name a file. And if we have multiple files with graph we need to mark all of them as an output.

kwchan1995:
> > Maybe I'm missing something - I thought everything you write using the filer will get packaged into the same jar, and therefore you should be able to retrieve them from within the [`shell command`](https://github.com/googleinterns/dagger-query/pull/17/files#diff-984f141cfa58d26ca608b4c2b15d488bR28)?
> 
> I think that the problem is here: [in this line](https://github.com/googleinterns/dagger-query/blob/d7a7815e1da67d12217bf6f6a33d0c62ed88701a/project/src/com/google/daggerquery/dagger_query_textproto.bzl#L42) we explicitly name a file. And if we have multiple files with graph we need to mark all of them as an output.

I see, what if we change the output to a zip file containing all the .textproto files? Then the consumer (presumably the UI) could unzip it afterwards?

margaritiko:
> > > Maybe I'm missing something - I thought everything you write using the filer will get packaged into the same jar, and therefore you should be able to retrieve them from within the [`shell command`](https://github.com/googleinterns/dagger-query/pull/17/files#diff-984f141cfa58d26ca608b4c2b15d488bR28)?
> > 
> > 
> > I think that the problem is here: [in this line](https://github.com/googleinterns/dagger-query/blob/d7a7815e1da67d12217bf6f6a33d0c62ed88701a/project/src/com/google/daggerquery/dagger_query_textproto.bzl#L42) we explicitly name a file. And if we have multiple files with graph we need to mark all of them as an output.
> 
> I see, what if we change the output to a zip file containing all the .textproto files? Then the consumer (presumably the UI) could unzip it afterwards?

I think this approach with .zip looks attractive. I am not sure I know how to extract zip from a client - should I implement another bazel rule?

kwchan1995:
> > > > Maybe I'm missing something - I thought everything you write using the filer will get packaged into the same jar, and therefore you should be able to retrieve them from within the [`shell command`](https://github.com/googleinterns/dagger-query/pull/17/files#diff-984f141cfa58d26ca608b4c2b15d488bR28)?
> > > 
> > > 
> > > I think that the problem is here: [in this line](https://github.com/googleinterns/dagger-query/blob/d7a7815e1da67d12217bf6f6a33d0c62ed88701a/project/src/com/google/daggerquery/dagger_query_textproto.bzl#L42) we explicitly name a file. And if we have multiple files with graph we need to mark all of them as an output.
> > 
> > 
> > I see, what if we change the output to a zip file containing all the .textproto files? Then the consumer (presumably the UI) could unzip it afterwards?
> 
> I think this approach with .zip looks attractive. I am not sure I know how to extract zip from a client - should I implement another bazel rule?

I think it's okay to let the client handle the unzipping however it likes. I don't think there are existing clients of the `dagger_query_textproto` rule [yet](https://github.com/googleinterns/dagger-query/search?q=dagger_query_textproto&unscoped_q=dagger_query_textproto). For example, if the zip file is passed to the client side UI, I'd imagine there would be a javascript library which could read zip files.

margaritiko:
> yet

Ok, I got it. In Java it will look like unzipping resource folder with using `ZipInputStream` etc. I marked this PR as a draft and will come back with updates later ✨

margaritiko:
@Ratchette @kwchan1995 @allie-wood

I've updated PR with new changes, in this one I supported saving binding graphs **in the plugin** and next pull request will update **the command line application** 🚀

project/src/com/google/daggerquery/dagger_query_textproto.bzl
```diff
@@ -25,7 +25,7 @@ def _dagger_query_textproto_impl(ctx):
     ctx.actions.run_shell(
         inputs = src_jars,
         outputs = [ctx.outputs.out],
-        command = "unzip -p {src_jar} binding_graph.textproto > {out}".format(
+        command = "unzip {src_jar} '*.textproto' -d .; zip -R {out} '*.textproto'".format(
```
kwchan1995: I think you don't need the `-R` (recursive) flag, as your `textproto` files are written to the top folder.

kwchan1995 (approved):
Awesome!

project/src/com/google/daggerquery/dagger_query_textproto.bzl
margaritiko: Yes, it looks unnecessary. I've just removed it ✔️

project/src/com/google/daggerquery/dagger_query_textproto.bzl
```diff
@@ -25,7 +25,7 @@ def _dagger_query_textproto_impl(ctx):
     ctx.actions.run_shell(
         inputs = src_jars,
         outputs = [ctx.outputs.out],
-        command = "unzip -p {src_jar} binding_graph.textproto > {out}".format(
+        command = "unzip {src_jar} '*.textproto' -d .; zip {out} '*.textproto'".format(
```
Ratchette: if it is common for a jar file to contain textprotos outside of this plugin, I think we should add a specific prefix or suffix to these binding graphs.

Ratchette (approved):

project/src/com/google/daggerquery/dagger_query_textproto.bzl
margaritiko: Yes, thanks! I think `_graph.textproto` suffix will be pretty good.

PR #19, opened 2020-08-19
Supported unzipping archive in a command-line app

margaritiko:
Supports unzipping archive with _.textproto files_ in a command-line app 📨

> The second step of supporting multiple graphs in a project. Plugin changes included in #17.

**Main changes:**
 1. Adds `guava` dependency for **QueryExecutor** and **SourcesLoader** as they use data structures and Files class from Guava.
 2. Supports analysis of multiple graphs in **QueryExecutor** by trying to execute a query in each of them.
 3. Supports extracting .zip file in **SourcesLoader** and deserializing binding graphs from all files located in .zip archive.

[Card](https://github.com/googleinterns/dagger-query/projects/1#card-43510033)

### DIFF
--- a/project/src/com/google/daggerquery/executor/QueryExecutor.java
+++ b/project/src/com/google/daggerquery/executor/QueryExecutor.java
@@ -16,9 +16,10 @@ limitations under the License.
 
 package com.google.daggerquery.executor;
 
+import com.google.common.collect.ImmutableList;
 import com.google.daggerquery.executor.models.Query;
 import com.google.daggerquery.executor.services.SourcesLoader;
-
+import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.util.Arrays;
@@ -43,8 +44,26 @@ public class QueryExecutor {
       }
 
       Query query = new Query(args[0], Arrays.copyOfRange(args, 1, args.length));
-      List<String> result = query.execute(new SourcesLoader().loadBindingGraph());
-      result.forEach(printStream::println);
+      List<BindingGraph> bindingGraphs = new SourcesLoader().loadBindingGraphs();
+
+      ImmutableList.Builder resultBuilder = new ImmutableList.Builder();
+
+      // We assume that we successfully executed a query only if in at least one graph it was executed without fail.
+      String failureReason = "";
+      for (BindingGraph bindingGraph: bindingGraphs) {
+        try {
+          resultBuilder.addAll(query.execute(bindingGraph));
+        } catch (IllegalArgumentException e) {
+          failureReason = e.getMessage();
+        }
+      }
+
+      ImmutableList resultList = resultBuilder.build();
+      if (resultList.isEmpty()) {
+        throw new IllegalArgumentException(failureReason);
+      } else {
+        resultList.forEach(printStream::println);
+      }
     } catch (IllegalArgumentException e) {
       printStream.println("Execution failed. Reason: " + e.getMessage());
     } catch (IOException e) {
--- a/project/src/com/google/daggerquery/executor/services/SourcesLoader.java
+++ b/project/src/com/google/daggerquery/executor/services/SourcesLoader.java
@@ -33,7 +33,7 @@ import java.util.zip.ZipFile;
  */
 public class SourcesLoader {
   private static final String PATH_TO_BINDING_GRAPHS = "/com/google/daggerquery/binding_graph_data.zip";
-  private static final String BINDING_GRAPHS_SOURCES = "binding_graphs.bin";
+  private static final String BINDING_GRAPHS_SOURCES = "binding_graphs.tmp";
 
   /**
    * Reads .zip resource file which contains several .textproto files with serialized binding graphs.
--- a/project/src/com/google/daggerquery/executor/services/SourcesLoader.java
+++ b/project/src/com/google/daggerquery/executor/services/SourcesLoader.java
@@ -16,24 +16,73 @@ limitations under the License.
 
 package com.google.daggerquery.executor.services;
 
+import com.google.common.io.Files;
 import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph;
+import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
 
 /**
  * A class which loads a binding graph saved with Dagger SPI plugin.
  */
 public class SourcesLoader {
-  private static final String PATH_TO_BINDING_GRAPH = "/com/google/daggerquery/binding_graph_data.textproto";
+  private static final String PATH_TO_BINDING_GRAPHS = "/com/google/daggerquery/binding_graph_data.zip";
+  private static final String BINDING_GRAPHS_SOURCES = "binding_graphs.bin";
 
-  public BindingGraph loadBindingGraph() throws IOException {
-    InputStream inputStream = SourcesLoader.class.getResourceAsStream(PATH_TO_BINDING_GRAPH);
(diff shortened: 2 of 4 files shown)


project/src/com/google/daggerquery/executor/QueryExecutor.java
```diff
@@ -43,8 +44,26 @@ public static void main(String[] args) {
...
+
+      ImmutableList.Builder resultBuilder = new ImmutableList.Builder();
+
+      // We assume that we successfully executed a query only if in at least one graph it was executed without fail.
+      String failureReason = "";
+      for (BindingGraph bindingGraph: bindingGraphs) {
+        try {
+          resultBuilder.addAll(query.execute(bindingGraph));
+        } catch (IllegalArgumentException e) {
+          failureReason = e.getMessage();
```
kwchan1995, line 57: This variable will get overwritten if there are multiple failures. Is that the intention?

kwchan1995: If you just care about *any* exception, how about just let the exception propagate up the call stack, as I can see you already have an outermost catch block on line 67?

project/src/com/google/daggerquery/executor/services/SourcesLoader.java
```diff
@@ -16,24 +16,73 @@
...
+  public List<BindingGraph> loadBindingGraphs() throws IOException {
+    try (InputStream zipInputStream = SourcesLoader.class.getResourceAsStream(PATH_TO_BINDING_GRAPHS)) {
+      if (zipInputStream == null) {
+        throw new FileNotFoundException(String.format("File %s is missing.", PATH_TO_BINDING_GRAPHS));
+      }
 
-    if (inputStream == null) {
-      throw new FileNotFoundException(String.format("File %s is missing.", PATH_TO_BINDING_GRAPH));
+      ZipFile zipFile = makeZipFileFromInputStream(zipInputStream);
+      Enumeration<? extends ZipEntry> filesWithBindingGraphs = zipFile.entries();
```
kwchan1995: Instead of having to define your helper method, would looping directly over [`ZipInputStream.getNextEntry()`](https://docs.oracle.com/javase/8/docs/api/java/util/zip/ZipInputStream.html#getNextEntry--) work?

project/src/com/google/daggerquery/executor/QueryExecutor.java
margaritiko: If we will ignore this exception catch statement in a line 67 will get it, yes, but a query execution will stop after it. There is might be a situation when we have, for example, **three graphs**, in **two** of them **exception** will be thrown and in **one** of them, the query will be executed **correctly** (because there is no intersection between bindings).

I thought that saving only one (the latest) error message is enough, the user can see others after fixing that one. This exception always points to the missing source node.

kwchan1995: > If we will ignore this exception catch statement in a line 67 will get it, yes, but a query execution will stop after it.

Ah you're right, thanks! :sweat_smile:

project/src/com/google/daggerquery/executor/services/SourcesLoader.java
margaritiko: I thought about it but I didn't find another easier way to read the content of .textproto files from .zip. For it now I use the method `getInputStream(...)` from `ZipFile` instance.

kwchan1995: Sorry maybe I wasn't clear - the `makeZipFileFromInputStream` helper method seems to just convert the `ZipInputStream` into `ZipFile`, and then `zipFile.entries()` is called to obtain `Enumeration<? extends ZipEntry>`. Based on the type signature I would have thought iterating through `ZipInputStream.getNextEntry()` would be equivalent to `Iterable<ZipEntry>`. I haven't tried this myself so I could be wrong...

margaritiko: Yes, I got it. I can use it for getting entries this way: 

```
ZipEntry bindingGraphEntry = zipInputStream.getNextEntry();
while (bindingGraphEntry != null) {

    try (InputStream bindingGraphInputStream = zipFile.getInputStream(bindingGraphEntry)) {
        bindingGraphs.add(BindingGraph.parseFrom(bindingGraphInputStream));
    }

    bindingGraphEntry = zipInputStream.getNextEntry();
}
```

The problem is that now I use `ZipFile` instance also in a line 59, that's why it's hard to get rid of this helper method.

kwchan1995: Oh I see, yeah you taught me something new :smiley:

kwchan1995 (approved):

project/src/com/google/daggerquery/executor/services/SourcesLoader.java
```diff
@@ -16,24 +16,73 @@
...
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
 
 /**
  * A class which loads a binding graph saved with Dagger SPI plugin.
  */
 public class SourcesLoader {
-  private static final String PATH_TO_BINDING_GRAPH = "/com/google/daggerquery/binding_graph_data.textproto";
+  private static final String PATH_TO_BINDING_GRAPHS = "/com/google/daggerquery/binding_graph_data.zip";
+  private static final String BINDING_GRAPHS_SOURCES = "binding_graphs.bin";
```
Ratchette: Nit: 
It looks like bin is a commonly used binary format
https://www.reviversoft.com/file-extensions/bin

Since it looks like this file is only used as a temporary file to construct the zip file, can we use .tmp instead?

project/src/com/google/daggerquery/executor/QueryExecutor.java
```diff
@@ -43,8 +44,26 @@ public static void main(String[] args) {
...
+
+      ImmutableList resultList = resultBuilder.build();
+      if (resultList.isEmpty()) {
+        throw new IllegalArgumentException(failureReason);
+      } else {
+        resultList.forEach(printStream::println);
+      }
     } catch (IllegalArgumentException e) {
       printStream.println("Execution failed. Reason: " + e.getMessage());
     } catch (IOException e) {
```
Ratchette, line 69: If the ZIP archive contains a file that is not a Binding Graph textproto, BindingGraph.parseFrom() (Line 60 in SourcesLoader.java) will throw a InvalidProtocolBufferException.
https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Parser.html#parseFrom-java.io.InputStream-
This looks to be a subclass of an IOException, and so will hit this block here.

Can we add a block to catch InvalidProtocolBufferException and let the user know which file was not formatted correctly?

Ratchette (approved):

project/src/com/google/daggerquery/executor/QueryExecutor.java
margaritiko: Maybe it makes sense to catch this exception in `SourceLoader` and ignore it? What if the user has .textproto files and one non-textproto that appeared in .zip for some reason, then we can still parse the graph from .textproto and ignore all other files.

project/src/com/google/daggerquery/executor/services/SourcesLoader.java
margaritiko: Yes, makes sense, thank you! This .tmp extension better conveys the meaning of the content.

margaritiko:
In the latest commit I am removing a .tmp file after reading all content from it. Also `ZipFile` is Closable so I use `try-with-resources` statement to open it.

project/src/com/google/daggerquery/executor/QueryExecutor.java
Ratchette: That's a great solution. 

Could you also add an exception in loadBindingGraphs if we find 0 binding graphs?

margaritiko: Done ✅ I throw `FileNotFoundException` if `bindingGraphs` list is empty after reading all files.

PR #20, opened 2020-08-19
Adds Javadoc where it was missing

margaritiko:
Improves Javadoc by adding more details

**Main changes:**
 1. Adds Javadoc to public method `visitGraph(...)` in `QueryPlugin`.
 2. Adds `@throws` tags for methods which can throw an exception.
 3. Highlights the sense of `findAllPaths(...)` method.


Ratchette (approved):

kwchan1995 (approved):
Well done!

PR #21, opened 2020-08-19
Supported SOMEPATH query 🔫

margaritiko:
Supports SOMEPATH query

**Details:**
* _somepath_ implementation is deterministic and will return the same path for the same input
* _somepath_ works faster than _allpaths_ because it stops after it finds any path between `source` and `target` nodes

**Alternatives:**
Query _somepath_ can return the shortest path but in order to find it we will execute `findAllPaths(...)` method and then sort an output - it's quite long and may lead to big delays in a huge graph. The main purpose of _somepath_ is to figure out if there are any paths between nodes at all and to show any of them, so it is preferable to return the first path found by DFS. This approach is consistent with _somepath_ implementation in [bazel query language](https://github.com/bazelbuild/bazel/blob/37f11bfd2dc528fcf61b9162b277c7355e926396/src/main/java/com/google/devtools/build/lib/query2/query/PathLabelVisitor.java#L79).

**Main changes:**
1. Registers new _somepath_ query type in a `Query` class.
2. In `Query` class implements method `findSomePath(...)` which is used in the public method `execute(...)` to support _somepath_ query. This method contains a slightly modified implementation of Depth-first search algorithm.
3. Covered query execution algorithm with tests in `QueryTest`. 

[Card №1](https://github.com/googleinterns/dagger-query/projects/1#card-43792823)
[Card №2](https://github.com/googleinterns/dagger-query/projects/1#card-43792840)

### DIFF
--- a/project/src/com/google/daggerquery/executor/models/Query.java
+++ b/project/src/com/google/daggerquery/executor/models/Query.java
@@ -22,6 +22,7 @@ import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph;
 import com.google.daggerquery.protobuf.autogen.DependencyProto.Dependency;
 import sun.reflect.generics.reflectiveObjects.NotImplementedException;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Deque;
 import java.util.HashSet;
 import java.util.List;
@@ -42,6 +43,7 @@ public class Query {
 
   private final static String DEPS_QUERY_NAME = "deps";
   private final static String ALLPATHS_QUERY_NAME = "allpaths";
+  private final static String SOMEPATH_QUERY_NAME = "somepath";
 
   /**
    * The key is the name of supported query and the value is the number of parameters,
@@ -50,6 +52,7 @@ public class Query {
   private final static ImmutableMap<String, Integer> supportedQueries = ImmutableMap.<String, Integer>builder()
       .put(DEPS_QUERY_NAME, 1)
       .put(ALLPATHS_QUERY_NAME, 2)
+      .put(SOMEPATH_QUERY_NAME, 2)
       .build();
 
   private String name;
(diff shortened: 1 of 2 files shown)


project/tests/com/google/daggerquery/executor/models/QueryTest.java
```diff
@@ -202,6 +204,100 @@ public void testExecutingAllPathsQuery_WithNullBindingGraph_ThrowsNullPointerExc
...
+    Set<String> possibleOutputs = Sets.newHashSet(
+        "com.google.Component -> com.google.Details",
+        "com.google.Component -> com.google.Cat -> com.google.Details",
+        "com.google.Component -> com.google.CatsFactory -> com.google.Cat -> com.google.Details",
+        "com.google.Component -> com.google.Helper -> com.google.CatsFactory -> com.google.Cat -> com.google.Details"
+    );
+
+
+    assertEquals(1, queryExecutionResult.size());
+    assertTrue(possibleOutputs.contains(queryExecutionResult.get(0)));
```
Ratchette, line 262: This was quite a smart decision. 
It makes it clear that you cannot make any assumptions as to which path will be chosen.

Ratchette (approved):

project/tests/com/google/daggerquery/executor/models/QueryTest.java
margaritiko: Yes exactly! 🚀

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -99,16 +103,17 @@ public Query(String typeName, String... parameters) {
...
-      case DEPS_QUERY_NAME:
-        String sourceNode = parameters[0];
+      case DEPS_QUERY_NAME: {
+        String source = parameters[0];
 
-        if (!bindingGraph.getAdjacencyListMap().containsKey(sourceNode)) {
-          throw new IllegalArgumentException(String.format("Specified source node %s doesn't exist.", sourceNode));
+        if (!bindingGraph.getAdjacencyListMap().containsKey(source)) {
+          throw new IllegalArgumentException(String.format("Specified source node %s doesn't exist.", source));
         }
```
kwchan1995, line 111: nit: indentation seems off

kwchan1995 (approved):

PR #22, opened 2020-08-20
Changed command to avoid including files from current directory

margaritiko:
Changes command shell by removing using a wildcard in zip command

I realized that it's possible to use wildcards to archive files of **specific extensions only** so `*_graph.textproto` will not work properly. Also with this implementation using flag `-r` is required because without it `zip` command fails to pack everything in current `.` directory.

### DIFF
--- a/project/src/com/google/daggerquery/dagger_query_textproto.bzl
+++ b/project/src/com/google/daggerquery/dagger_query_textproto.bzl
@@ -25,7 +25,7 @@ def _dagger_query_textproto_impl(ctx):
     ctx.actions.run_shell(
         inputs = src_jars,
         outputs = [ctx.outputs.out],
-        command = "unzip {src_jar} '*_graph.textproto' -d .; zip {out} '*_graph.textproto'".format(
+        command = "unzip {src_jar} '*_graph.textproto' -d graphs; cd graphs; zip -r ../{out} .".format(
             src_jar = src_jars[0].path,
             out = ctx.outputs.out.path,
         ),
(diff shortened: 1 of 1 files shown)


project/src/com/google/daggerquery/dagger_query_textproto.bzl
```diff
@@ -25,7 +25,7 @@ def _dagger_query_textproto_impl(ctx):
     ctx.actions.run_shell(
         inputs = src_jars,
         outputs = [ctx.outputs.out],
-        command = "unzip {src_jar} '*_graph.textproto' -d .; zip {out} '*_graph.textproto'".format(
+        command = "unzip {src_jar} '*_graph.textproto' -d graphs; cd graphs; zip -r ../{out} .".format(
```
Ratchette: Nit: would you mind making this a bit more descriptive?
eg. binding_graphs_tmp

Ratchette: Would you mind also deleting this temporary directory after the binding graph zipfile has been created?

Ratchette (commented):
I love reviewing small changes 😀

project/src/com/google/daggerquery/dagger_query_textproto.bzl
margaritiko: Thanks, good idea ✨✔️

margaritiko: Done ✅

Ratchette (approved):

kwchan1995 (approved):

PR #23, opened 2020-08-20
Added instructions for users 💻

margaritiko:
Adds instructions which can help to set up a project

**Main changes:**
1. `README` updated with three steps required to start using Dagger Query project.
2. Adds `dagger-query.sh` which hides `bazel run` command from a user and executes queries.
3. Specifies `resources` attribute in the command-line app BUILD file.
4. Shows how to use `dagger_query_textproto` rule.

[Card](https://github.com/googleinterns/dagger-query/projects/1#card-43589741)


project/dagger-query.sh
```diff
@@ -0,0 +1,7 @@
+export PATH="$PATH:$HOME/bin"
+
+if [[ -z $3 ]]; then
```
kwchan1995: Would it be possible to pass through user arguments with `$@`?

project/src/com/google/daggerquery/BUILD
```diff
@@ -13,10 +13,20 @@
...
     deps = [
         "//src/com/google/daggerquery/plugin:plugin_sources",
     ]
 )
+
+dagger_query_textproto(
+    name = "binding_graph_data",
+
+    # Put here the name of a target in which you have used a plugin for generating binding graph.
+    dagger_app_target = "YOUR_APP_TARGET"
```
kwchan1995: Well I'm not sure if it's a good idea to check this in because this target will not build. How about moving this example to `README.md`?

project/src/com/google/daggerquery/executor/BUILD
```diff
@@ -24,4 +24,7 @@ java_binary(
         "//src/com/google/daggerquery/executor/models:query_executor_models",
         "//src/com/google/daggerquery/executor/services:query_executor_services",
     ],
+    resources = [
+        "//src/com/google/daggerquery:binding_graph_data"
```
kwchan1995: I'm not sure it makes sense to hardcode the textproto to `query_executor`. How about creating a macro (like the `dagger_query_textproto`) so that this dependency can be specified by the user via arguments?

project/dagger-query.sh
margaritiko: Yeah, that works! Thanks, I didn't know about it before.

project/src/com/google/daggerquery/executor/BUILD
margaritiko: Yes, it might make sense not to hardcode it here. However I think I didn't understand how this macro will be used - you mean user don't need to specify anything in `resources` at all by himself?

project/src/com/google/daggerquery/BUILD
margaritiko: Done! I've updated second step in README with this rule and removed it from this `BUILD` file.

project/src/com/google/daggerquery/executor/BUILD
kwchan1995: I was thinking if the user could just call your macro from their project's `BUILD` file, e.g.

```
# user's BUILD file
dagger_query(
  name = "app_dagger_query",
  dagger_app_target = "MY_APP_TARGET"
)
```

and then in your implementation:

```
# your macro implementation that accepts user arguments
def dagger_query(name, dagger_app_target):
   dagger_query_textproto(
       name = "binding_graph_data",
       dagger_app_target = "YOUR_TARGET"
  )

  java_binary(
     name = name,
     exports = ["//src/com/google/daggerquery/executor:query_executor"],
     resources = [":binding_graph_data"],
  )
```

and your `dagger-query.sh` script may need to take `//...:app_dagger_query` as a user supplied argument as well.

margaritiko: I got it, thanks. I've added this rule in the latest commit.

However `java_binary` doesn't support `exports` argument and maybe [my realisation of `dagger_query` rule](https://github.com/googleinterns/dagger-query/blob/15b80bc4b218e02eff8f89a312523f6087e158c1/project/src/com/google/daggerquery/executor/dagger_query.bzl#L24) is a little heavy because of this?

project/src/com/google/daggerquery/executor/dagger_query.bzl
```diff
@@ -0,0 +1,30 @@
...
+
+def dagger_query(name, dagger_app_target):
+   dagger_query_textproto(
+       name = "binding_graph_data",
+       dagger_app_target = dagger_app_target
+   )
+
+   java_binary(
+     name = name,
+     srcs = ["//src/com/google/daggerquery/executor:QueryExecutor.java"],
```
kwchan1995: Hmm... it feels weird to have the srcs defined in the macro, because this macro is effectively expanded in the user's `BUILD` file. Would it be possible to keep the original `:query_executor` target in `src/com/google/daggerquery/executor/BUILD` and reference it from here via the `runtime_deps` attribute?

margaritiko: I've tried this:
```
def dagger_query(name, dagger_app_target):
   dagger_query_textproto(
       name = "binding_graph_data",
       dagger_app_target = dagger_app_target
   )

   java_binary(
     name = name,
     runtime_deps = ["//src/com/google/daggerquery/executor:query_executor"],
     resources = [":binding_graph_data"],
   )
```
But without specifying `main_class` it complains about it with `ClassNotFoundException`: **Could not find or load main class com.google.daggerquery.dagger_query_app**. 

I can pass main class `main_class = "com.google.daggerquery.executor.QueryExecutor"` and get `ClassNotFoundException` caused by not being able to find the Query class (because `runtime_deps` is not enough and we need these dependencies at compile time).

Please, check new commit now, I've tried another approach and removed `java_binary` from executor/BUILD.

kwchan1995: Ah, my bad, I looked at https://docs.bazel.build/versions/master/be/java.html#java_binary again and there's a snippet at the bottom:

> The following code snippet illustrates a common mistake:
```
 java_binary(
    name = "DontDoThis",
    srcs = [
        ...,
        "GeneratedJavaFile.java",  # a generated .java file
    ],
    deps = [":generating_rule",],  # rule that generates that file
)
```
> Do this instead:
```
java_binary(
    name = "DoThisInstead",
    srcs = [
        ...,
        ":generating_rule",
    ],
)
```

so I wonder if we could do:

```
java_binary(
  name  = name,
  srcs = ["//src/com/google/daggerquery/executor:query_executor"],
  resources = [":binding_graph_data"],
)
```
and keep the original `:query_executor` rule?

Apologies for the inconvenience, if it still doesn't work let's go with your solution for now :smiley:

margaritiko: It fails again with this reason:  `'//src/com/google/daggerquery/executor:query_executor' does not produce any java_binary srcs files (expected .java, .srcjar or .properties). Since this rule was created by the macro 'dagger_query', the error might have been caused by the macro implementation.`

It's strange, I thought that `java_binary` is a generating rule 🤔

kwchan1995: Ah you might need to change `//src/com/google/daggerquery/executor:query_executor` to `java_library` because it becomes just another target depended from here. Anyway let's not worry too much about this, we can revisit this later if we have time :smiley:

kwchan1995 (approved):

README.md
```diff
@@ -4,6 +4,25 @@ A graph based query language on top of [dagger](https://github.com/google/dagger
 
 Dagger Query can be used to investigate the dagger dependency graph.
 
+
+## How to use Dagger Query project?
+
+1. Add `//src/com/google/daggerquery:dagger_query_plugin` target in a **plugins** attribute of your app's target.
+2. Go to `src/com/google/daggerquery/BUILD` and use **dagger_query** rule with your app's target name.
```
Ratchette: Should they be modifying the dagger_query rule inside project/src/com/google/daggerquery/BUILD?
or should they be using this rule in their own codebase?

Ratchette (approved):
LGTM 🦆

project/src/com/google/daggerquery/executor/BUILD
margaritiko: (The question was discussed below, resolving this conversation, since all the changes were made in the last commits.) ✅

README.md
margaritiko: No, now it can be used anywhere! Thanks, I updated the second step in the README in the latest commit 😊

project/dagger-query.sh
```diff
@@ -0,0 +1,3 @@
+export PATH="$PATH:$HOME/bin"
+
+bazel run --sandbox_debug --sandbox_block_path=/usr/local //src/com/google/daggerquery/executor:query_executor $@
```
kwchan1995: Should `//src/com/google/daggerquery/executor:query_executor` be changed to reference user's `dagger_query` macro now?
If this is specified as the first parameter of the user input, then the `$@` will have it and we can simply remove `//src/com/google/daggerquery/executor:query_executor`.

kwchan1995: Please also update the `README.md` instructions afterwards.

margaritiko: Done 💫 [I've put these instructions](https://github.com/googleinterns/dagger-query/pull/23/commits/add4e702142aba60fdb59aa86418f414a3773754) in the third step in a README.

PR #24, opened 2020-08-26
Supported typos fixer ⚙️

margaritiko:
Adds an additional check for typos in the name of the binding node if the node was not found in the graph.

**Problem:** usually binding node names quite long and it may be difficult to write them correctly from scratch. This PR adds special check for typos in a name if the node was not found as a key in an adjacency list. 

> `deps com.goooogle.Beach`
**Result:** `Execution failed. Reason: Binding with name deps com.goooogle.Beach contains a typo and not found in a graph. Maybe you meant deps com.google.Beach?`

For measuring distances between strings uses [Levenshtein Distance](https://en.wikipedia.org/wiki/Levenshtein_distance) metric from [org.apache.commons.text.similarity](https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/similarity/LevenshteinDistance.html). The name is considered to contain a typo if the distance to the existing node is no more than 3.

**Main changes:**
 * In `WORKSPACE` file loads Apache Commons Text library jar.
 * This dependency is used in a `//src/com/google/daggerquery/executor/models/BUILD` file.
 * Implemented an unchecked `MisspelledNodeNameException` exception to determine when a typo was encountered while executing a request.
 * Added tests for each type of query, checking for situations when the node name contains a typo and when the number of corrections is too large to consider the names similar.

### DIFF
--- a/project/src/com/google/daggerquery/executor/QueryExecutor.java
+++ b/project/src/com/google/daggerquery/executor/QueryExecutor.java
@@ -17,6 +17,7 @@ limitations under the License.
 package com.google.daggerquery.executor;
 
 import com.google.common.collect.ImmutableList;
+import com.google.daggerquery.executor.models.MisspelledNodeNameException;
 import com.google.daggerquery.executor.models.Query;
 import com.google.daggerquery.executor.services.SourcesLoader;
 import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph;
@@ -49,18 +50,25 @@ public class QueryExecutor {
       ImmutableList.Builder resultBuilder = new ImmutableList.Builder();
 
       // We assume that we successfully executed a query only if in at least one graph it was executed without fail.
-      String failureReason = "";
+      MisspelledNodeNameException lastMisspelledNodeNameException = null;
+      IllegalArgumentException lastIllegalArgumentException = null;
       for (BindingGraph bindingGraph: bindingGraphs) {
         try {
           resultBuilder.addAll(query.execute(bindingGraph));
         } catch (IllegalArgumentException e) {
-          failureReason = e.getMessage();
+          lastIllegalArgumentException = e;
+        } catch (MisspelledNodeNameException e) {
+          lastMisspelledNodeNameException = e;
         }
       }
 
       ImmutableList resultList = resultBuilder.build();
       if (resultList.isEmpty()) {
-        throw new IllegalArgumentException(failureReason);
+        if (lastMisspelledNodeNameException != null) {
+          throw new IllegalArgumentException(lastMisspelledNodeNameException.getMessage());
+        } else if (lastIllegalArgumentException != null) {
+          throw lastIllegalArgumentException;
+        }
       } else {
         resultList.forEach(printStream::println);
       }
--- a/project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
+++ b/project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
@@ -1,0 +1,21 @@
+package com.google.daggerquery.executor.models;
+
+/**
+ * An exception is thrown to indicate that a binding name contains a typo.
+ *
+ * <p>This exception <b>should</b> be used to report a typo in a binding node name,
+ * which user has provided as a query parameter.
+ */
+public class MisspelledNodeNameException extends RuntimeException {
+
+  /**
+   * Constructs a <code>MisspelledNodeNameException</code> with a detail message
+   * consisting of the given binding name with typo followed by the correct binding name
+   * which presented in a binding graph.
+   */
+  public MisspelledNodeNameException(String nodeNameWithTypo, String correctNodeName) {
+    super(String.format("Binding with name %s contains a typo and not found in a graph. Maybe you meant %s?",
+                        nodeNameWithTypo, correctNodeName));
+  }
+
+}
--- a/project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
+++ b/project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
@@ -9,7 +9,7 @@ package com.google.daggerquery.executor.models;
 public class MisspelledNodeNameException extends RuntimeException {
 
   /**
-   * Constructs a <code>MisspelledNodeNameException</code> with a detail message
+   * Constructs a <code>MisspelledNodeNameException</code> with a detailed message
    * consisting of the given binding name with typo followed by the correct binding name
    * which presented in a binding graph.
    */
--- a/project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
+++ b/project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
@@ -1,5 +1,7 @@
 package com.google.daggerquery.executor.models;
 
+import java.util.List;
+
 /**
  * An exception is thrown to indicate that a binding name contains a typo.
  *
@@ -10,12 +12,12 @@ public class MisspelledNodeNameException extends RuntimeException {
 
   /**
    * Constructs a <code>MisspelledNodeNameException</code> with a detailed message
-   * consisting of the given binding name with typo followed by the correct binding name
-   * which presented in a binding graph.
+   * consisting of the provided binding name, followed by similarly named bindings
+   * that exist in our binding graph.
    */
-  public MisspelledNodeNameException(String nodeNameWithTypo, String correctNodeName) {
+  public MisspelledNodeNameException(String nodeNameWithTypo, List<String> correctNodesName) {
     super(String.format("Binding with name %s contains a typo and not found in a graph. Maybe you meant %s?",
-                        nodeNameWithTypo, correctNodeName));
+                        nodeNameWithTypo, String.join(" or ", correctNodesName)));
   }
 
 }
--- a/project/src/com/google/daggerquery/executor/models/Query.java
+++ b/project/src/com/google/daggerquery/executor/models/Query.java
@@ -306,15 +306,10 @@ public class Query {
    */
   private List<String> findNodesWithClosestName(String originalNode, Set<String> allNodes) {
     List<String> closestNodes = new ArrayList<>();
-    int bestDistance = Integer.MAX_VALUE;
 
     for (String node: allNodes) {
       int distance = levenshteinDistanceCalculator.apply(originalNode, node);
-      if (distance <= MAX_NUMBER_OF_MISPLACED_LETTERS && distance < bestDistance) {
-        closestNodes.clear();
-        closestNodes.add(node);
-        bestDistance = distance;
-      } else if (distance == bestDistance) {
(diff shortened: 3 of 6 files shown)


project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * An exception is thrown to indicate that a binding name contains a typo.
+ *
+ * <p>This exception <b>should</b> be used to report a typo in a binding node name,
+ * which user has provided as a query parameter.
+ */
+public class MisspelledNodeNameException extends RuntimeException {
+
+  /**
+   * Constructs a <code>MisspelledNodeNameException</code> with a detail message
```
Ratchette: Nit: 
with a **detailed** message

project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
```diff
@@ -0,0 +1,21 @@
...
+   * consisting of the given binding name with typo followed by the correct binding name
+   * which presented in a binding graph.
```
Ratchette: Are you thinking along the lines that this class will find one or more nodes in the binding graph with similar names?
If so, then we might be able to rephrase this to something like:
"... the provided binding name, followed by similarly named bindings that exist in our binding graph." 

eg. if in our graph we have bindings like Cat or Cab, and the User inputs Can, it might be difficult to determine which is the correct binding / the binding they were thinking of.

project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
```diff
@@ -0,0 +1,21 @@
...
+   */
+  public MisspelledNodeNameException(String nodeNameWithTypo, String correctNodeName) {
```
Ratchette: Can we change this to a list of strings? (eg. if there are a few close possibilities)

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -99,26 +107,25 @@ public Query(String typeName, String... parameters) {
...
    */
   public List<String> execute(BindingGraph bindingGraph) {
     switch (name) {
       case DEPS_QUERY_NAME: {
         String source = parameters[0];
 
-        if (!bindingGraph.getAdjacencyListMap().containsKey(source)) {
-          throw new IllegalArgumentException(String.format("Specified source node %s doesn't exist.", source));
-        }
+        checkNodeForCorrectness(source, bindingGraph);
```
Ratchette, line 120: Since it looks like we're checking in every case of the switch statement with the same arguments, can we move this to be above the switch statement?

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -260,4 +265,56 @@ private boolean findSomePath(String source, String target, Path path,
 
     return false;
   }
+
+  /**
+   * Checks if the passed {@code node} is on the {@link BindingGraph} or if the user misspelled the node's name.
```

... node} is **in** the {@link ...

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -46,6 +47,13 @@
   private final static String ALLPATHS_QUERY_NAME = "allpaths";
   private final static String SOMEPATH_QUERY_NAME = "somepath";
 
+  private final static int MAX_NUMBER_OF_MISPLACED_LETTERS = 3;
```
Ratchette, line 50: Out of curiosity, how did you settle on a maximum of 3?

Ratchette (changes requested):

project/src/com/google/daggerquery/executor/QueryExecutor.java
```diff
@@ -49,18 +50,25 @@ public static void main(String[] args) {
       ImmutableList.Builder resultBuilder = new ImmutableList.Builder();
 
       // We assume that we successfully executed a query only if in at least one graph it was executed without fail.
-      String failureReason = "";
+      MisspelledNodeNameException lastMisspelledNodeNameException = null;
+      IllegalArgumentException lastIllegalArgumentException = null;
```
kwchan1995, line 54: This pattern doesn't look very sustainable as more exception types are added. How about just keeping a single variable `Exception e` if you only care about any exception?

project/tests/com/google/daggerquery/executor/models/QueryTest.java
```diff
@@ -111,6 +111,28 @@ public void testExecutingDepsQuery_WithNullBindingGraph_ThrowsNullPointerExcepti
...
+      List<String> queryExecutionResult = query.execute(makeSimpleBindingGraph());
+      fail();
+    } catch (MisspelledNodeNameException e) {
+      // We do not care too much about the rest of the message, but it should contain a correct node name.
+      assertTrue(e.getMessage().contains("com.google.CatsFactory"));
+    }
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testExecutingDepsQuery_WithFourTyposInNodeName_ThrowsIllegalArgumentException() {
```
kwchan1995: nit: How about changing `Four` to `TooMany` to make it more descriptive?

project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
margaritiko: Done ✅

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: Done ✅

project/tests/com/google/daggerquery/executor/models/QueryTest.java
margaritiko: Good idea, done ✅

project/src/com/google/daggerquery/executor/QueryExecutor.java
margaritiko: Yes, I'm not happy with my approach either, but keeping any `Exception` will also not work: `MisspelledNodeNameException` has a higher priority, and we want to show the user the fix, if any.

The second idea is to cast the last saved exception to  `MisspelledNodeNameException` and only rewrite it if the cast has failed. Is it better?

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: I thought that this is not too small and not too many number of fixes 🦄. Do you think it makes sense to make it more or less?

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: This will force us to take the first source parameter and pass it to the function, but in future queries, the source node may not be passed as the first parameter. 

We [discussed this here in a previous PR](https://github.com/googleinterns/dagger-query/pull/15/files/27d926ba84abfe8e00f5bfe9d6ca19c4acbf677d#r470689088) and decided it might be better to repeat some logic.

project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
margaritiko: Thanks for the idea, with this improvement the feature is really more useful! ✨🌟

project/src/com/google/daggerquery/executor/models/MisspelledNodeNameException.java
margaritiko: Great, thanks. I made [all the changes in the last commit](https://github.com/googleinterns/dagger-query/pull/24/commits/2b3a9d0cc5155f01689d7996f3cb34063c115419).

project/src/com/google/daggerquery/executor/QueryExecutor.java
kwchan1995: I see, thanks I was missing this earlier. How about just keeping a variable denoting the last seen exception, and break out of the loop once a `MisspelledNodeNameException` has been caught?

margaritiko: What if in the first graph we will catch `MisspelledNodeNameException` with a fix from `com.google.Cat` to `com.google.Can`, but in the second graph `com.google.Cat` node exists? It sounds like a very rare case, buuut 🤔

kwchan1995: That's true :joy: I'm not sure what would be the ideal way now - another idea I can think of is to collect all the errors in a list, then later have aggregation logic to combine these errors for display (e.g. sort the list by importance), but this sounds overkill and may not make the code any cleaner :disappointed: 
Let's do it your way until we have a better solution.

project/src/com/google/daggerquery/executor/models/Query.java
Ratchette: My gut is that it might need to be more, but I have no data to back this up. So let's leave it for now, and change it if we see any potential problems cropping up.

For more context, I'm a terrible speller. The most common mistake that I make is inversion (eg. thsi  instead of this). Using the LevenshteinDistance an inversion would mean that two letters are misplaced. So using the example of "thsi" above, LevenshteinDistance would say that "his" is a closer match than "this" although while typing I feel like the latter may be just as likely an error.

... But as you can see, this opinion is entirely anecdotal. So unless we feel like linguistically analyzing typing mistakes, I think that your initial number is a good place to start.

project/src/com/google/daggerquery/executor/models/Query.java
Ratchette: Thanks for linking that conversation, I defer to Karl's expertise :)

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -260,4 +265,60 @@ private boolean findSomePath(String source, String target, Path path,
...
+   * <p>If no nodes with a similar name are found, returns an empty {@link List<String>}.
+   * Otherwise, it returns a {@link List<String>} with nodes with the same distance.
+   */
+  private List<String> findNodesWithClosestName(String originalNode, Set<String> allNodes) {
+    List<String> closestNodes = new ArrayList<>();
+    int bestDistance = Integer.MAX_VALUE;
+
+    for (String node: allNodes) {
+      int distance = levenshteinDistanceCalculator.apply(originalNode, node);
+      if (distance <= MAX_NUMBER_OF_MISPLACED_LETTERS && distance < bestDistance) {
```
Ratchette: Now that we're looking for all of the nodes within 3 steps of the user input (vs the closest node) do we still need this best distance check?

Ratchette (changes requested):

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: Actually no, let's save all suggestions with the distance that works for us 👌

Ratchette (approved):
LGTM 🏄‍♂️

kwchan1995 (approved):

PR #25, opened 2020-08-26
Update README.md

Ratchette:
Move the disclaimer that this is not an officially supported Google product higher up to be more visible.


margaritiko (approved):
🚀

kwchan1995 (approved):

PR #26, opened 2020-08-31
Layout a page with an input field using Bootstrap 🍰

margaritiko:
Layout a page with an input field using Bootstrap.

**We decided to use [Bootstrap](https://getbootstrap.com/) because it makes layout easier and provides a lot of nice components. In this PR I have added customised input field for typing queries.**

[Card](https://github.com/googleinterns/dagger-query/projects/2#card-44227070)

**Problem:** The classic approach to supporting Bootstrap in our project is to use [Bootstrap CDN links](https://getbootstrap.com/docs/3.3/getting-started/#download-cdn) and have an internet connection, but the Dagger Query UI can run locally offline and we don't want to limit it. Another typical approach is to use [npm](https://getbootstrap.com/docs/3.3/getting-started/#download-npm), but it forces all Dagger Query UI users to install it on their machines, which is also bad.

**Solution:** I put all the Bootstrap required files locally in the _third-party_ folder. All of these files are under [the MIT license](https://opensource.org/licenses/MIT) and we can use them.

### Page with input field:

text-field


project/src/com/google/daggerqueryui/page.html
```diff
@@ -0,0 +1,39 @@
...
+    <!-- JQuery -->
+    <script src="third-party/js/jquery.min.js"></script>
+    <!-- MDB core JavaScript -->
+    <script type="text/javascript" src="third-party/js/mdb.min.js"></script>
+</head>
+<body>
+
+<div class="container" id="query-text-field">
+    <div class="md-form form-lg">
+        <input type="text" id="inputLGEx" class="form-control form-control-lg" spellcheck="false">
```
Ratchette: I'm not sure I understand what inputLGEx means. Is LGEx shorthand for a framework?

Ratchette (commented):
Could we rename the names of the html & css files to index?
eg. index.html and index.css?

Suraj may have better ideas on what we should call the "default" webpage.

project/src/com/google/daggerqueryui/page.html
margaritiko: Yep, it is widely used in all uses of the framework, but it makes no sense to use this id here so I have renamed it to `query-input` ✅

project/src/com/google/daggerqueryui/page.html
```diff
@@ -0,0 +1,39 @@
...
+    <link rel="stylesheet" href="third-party/css/roboto-fonts.css">
+
+    <!-- JQuery -->
+    <script src="third-party/js/jquery.min.js"></script>
+    <!-- MDB core JavaScript -->
+    <script type="text/javascript" src="third-party/js/mdb.min.js"></script>
+</head>
+<body>
+
+<div class="container" id="query-text-field">
```
o-y: It seems you're mixing the naming convention of the `id` attribute between hyphen-separated case (`"query-text-field"`) and camel case (`"inputLGEx"`). Generally it's preferable to keep the id attributes naming convention consistent with each other.

Ratchette (approved):
LGTM 🕸

project/src/com/google/daggerqueryui/index.html
```diff
@@ -0,0 +1,39 @@
...
+
+    <link rel="stylesheet" href="page-style.css">
+    <link rel="stylesheet" href="third-party/css/bootstrap.min.css">
+    <link rel="stylesheet" href="third-party/css/mdb.min.css">
+    <link rel="stylesheet" href="third-party/css/roboto-fonts.css">
+
+    <!-- JQuery -->
+    <script src="third-party/js/jquery.min.js"></script>
+    <!-- MDB core JavaScript -->
+    <script type="text/javascript" src="third-party/js/mdb.min.js"></script>
```
kwchan1995: nit: How about adding the `defer` attribute to prevent javascript download from blocking the page load? See https://stackoverflow.com/a/24070373

project/src/com/google/daggerqueryui/page.html
```diff
@@ -0,0 +1,39 @@
...
+
+    <link rel="stylesheet" href="page-style.css">
+    <link rel="stylesheet" href="third-party/css/bootstrap.min.css">
+    <link rel="stylesheet" href="third-party/css/mdb.min.css">
+    <link rel="stylesheet" href="third-party/css/roboto-fonts.css">
+
+    <!-- JQuery -->
+    <script src="third-party/js/jquery.min.js"></script>
+    <!-- MDB core JavaScript -->
+    <script type="text/javascript" src="third-party/js/mdb.min.js"></script>
```
o-y: Super nit: One script tag has the attribute `type="text/javascript"` and the other doesn't. That said, as JavaScript is the default programming language of the browser the attribute isn't needed.

project/src/com/google/daggerqueryui/third-party/css/roboto-fonts.css
```diff
@@ -0,0 +1,252 @@
+/* cyrillic-ext */
+@font-face {
+  font-family: 'Roboto';
+  font-style: normal;
+  font-weight: 300;
+  font-display: swap;
+  src: local('Roboto Light'), local('Roboto-Light'), url(https://fonts.gstatic.com/s/roboto/v20/KFOlCnqEu92Fr1MmSU5fCRc4AMP6lbBP.woff2) format('woff2');
```
kwchan1995, line 7: optional: Not sure if you want to make these offline as well if you're really concerned about internet availability. I wouldn't worry too much :stuck_out_tongue:

kwchan1995 (approved):

project/src/com/google/daggerqueryui/third-party/css/roboto-fonts.css
margaritiko: I see, thanks for noticing this. Boostrap's [default font family](https://www.w3schools.com/bootstrap4/bootstrap_typography.asp#:~:text=Bootstrap%204%20uses%20a%20default,1rem%20(16px%20by%20default).) is "Helvetica Neue" and it looks pretty good too. If there are no special requirements for using `Roboto` (it's quite popular in Google) I would rather remove it rather than download a lot of extra files.

project/src/com/google/daggerqueryui/index.html
margaritiko: Thanks for sharing, that's cool feature, I didn't know about `defer` before. I have added it ✅

project/src/com/google/daggerqueryui/page.html
margaritiko: Thanks! For consistency I'll use hyphen-separated case 👌

project/src/com/google/daggerqueryui/page.html
margaritiko: Thanks, I have removed it. In general, [as I see](https://stackoverflow.com/questions/21098865/text-javascript-vs-application-javascript) `text/javascript` is obsolete and people recommend to use `application/javascript` instead of it - are there any cases when it makes sense or specifying `javascript` in `type` attribute is always redundant?

project/src/com/google/daggerqueryui/third-party/css/roboto-fonts.css
margaritiko: Or we can leave everything as it is and use `Roboto` only if the Internet is available.

kwchan1995: Well it's just a prototype so I wouldn't worry too much about font requirements, let's just leave it aside :smiley:

project/src/com/google/daggerqueryui/page.html
o-y: Yep, if you're only targetting modern browsers then the `type` (MIME type) attribute is always redundant. The script type if inferred via [packet inspection](https://mimesniff.spec.whatwg.org/) by the browser (due to a lack of standardization). In HTML5 compliant browsers an omitted MIME type in a script tag indicates the script is JavaScript. As a result the spec urges authors to omit the attribute rather than provide a redundant MIME type.

LGTM!

PR #27, opened 2020-08-31
Highlights query name in the input field 🌱

margaritiko:
Highlights query name in the input field.

#### The idea is to highlight the query name if it is valid. This helps the user know that he is on the right way.

**Solution:** adds another input field with `width: auto` which is shown only when the first word in the input field is a valid query name. Script is written with [JQuery](https://jquery.com/).

### How does it look like?

highlight-query

[Card](https://github.com/googleinterns/dagger-query/projects/2#card-44227091)


Ratchette:
I've never created dependent changes in separate branches using git. 

Should this merge request be from highlight_query_name to master? 
It looks like this PR merges this branch (highlight_query_name) with the branch it is depending upon (adding_text_field). Does that mean you'll need to file a second PR to merge adding_text_field into master?

margaritiko:
> I've never created dependent changes in separate branches using git.
> 
> Should this merge request be from highlight_query_name to master?
> It looks like this PR merges this branch (highlight_query_name) with the branch it is depending upon (adding_text_field). Does that mean you'll need to file a second PR to merge adding_text_field into master?

This PR depends on the `add_text_field` branch because it requires some logic to be created in it. No other PR is needed, I will change the destination to master before the merge (after I will merge `add_text_field` PR, respectively).

project/src/com/google/daggerqueryui/scripts/highlight-query.js
```diff
@@ -0,0 +1,27 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+$(function (){
+    $.fn.validate_query = function (query){
+        return query == "deps" || query == "allpaths" || query == "somepath";
```
kwchan1995: Would it be possible to put supported query names in a reusable variable, so that we have a single source of truth and reduce maintenance?

project/src/com/google/daggerqueryui/scripts/highlight-query.js
```diff
@@ -0,0 +1,27 @@
...
+    };
+});
+
+$(document).on('keydown keyup change', '.interactive-input-container input', function (){
+    if(($(this).val().length) && ($(this).val().split(' ').length) && ($(this).validate_query($(this).val().split(' ')[0]))){
```
kwchan1995: How about avoiding repeated calls to `.val()` and `.split()` by storing intermediate variables?

project/src/com/google/daggerqueryui/page-style.css
```diff
@@ -15,3 +15,13 @@
...
     margin-top: 5%;
 }
+
+.interactive-input-container .query-name {
+    position: absolute;
+    width: auto;
+    display: none;
+    border: 0px;
+    color: #09AF00;
+    background: #FFF !important;
```
kwchan1995: nit: I think `white` is more readable than `#FFF`, but it may be a matter of taste.

kwchan1995 (approved):

Ratchette:
> > I've never created dependent changes in separate branches using git.
> > Should this merge request be from highlight_query_name to master?
> > It looks like this PR merges this branch (highlight_query_name) with the branch it is depending upon (adding_text_field). Does that mean you'll need to file a second PR to merge adding_text_field into master?
> 
> This PR depends on the `add_text_field` branch because it requires some logic to be created in it. No other PR is needed, I will change the destination to master before the merge (after I will merge `add_text_field` PR, respectively).

Great, thank you for the explanation.

project/src/com/google/daggerqueryui/scripts/highlight-query.js
Ratchette: Ditto. Would it be possible to format this file more like JavaScript? Eg. more intermediary variables, methods, even objects if need be.

project/src/com/google/daggerqueryui/scripts/highlight-query.js
o-y: After addressing @kwchan1995's comments; the if statement is still slightly hard to follow.
- It's not immediately obvious what the length comparisons are doing. To someone unfamilar to JS it looks like you're checking the existance of the length property, and not that the value is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy). 
- You could consider explictly ensuring the length is greater than 0 (`length > 0`), this removes any ambiguity.

project/src/com/google/daggerqueryui/scripts/highlight-query.js
```diff
@@ -0,0 +1,27 @@
...
+        $(this).closest('.interactive-input-container').find('.query-name').html($(this).val().split(' ')[0]).show();
+    } else{
```
o-y: Super nit: add a space after "else"

project/src/com/google/daggerqueryui/scripts/highlight-query.js
margaritiko: Thanks for suggestions! ✨

I have decided to remove these checks for length because according to the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#Using_split):
> When the string is empty, split() returns an array containing one empty string, rather than an empty array. This gives a simple solution - let's split first and then check the first element from the array.

project/src/com/google/daggerqueryui/scripts/highlight-query.js
margaritiko: Done ✅

Also I have added some spaces after `if` - is there any specific code style for jQuery scripts or I can use JavaScript conventions?

project/src/com/google/daggerqueryui/scripts/highlight-query.js
margaritiko: Thanks, done ✅🚀

project/src/com/google/daggerqueryui/scripts/highlight-query.js
margaritiko: Yep, I have put queries names into array and then checked with `.includes(...)` ✔️

project/src/com/google/daggerqueryui/page-style.css
margaritiko: At first I tried different hex values for the color, but then I decided to stick with white - let's replace the hex name with `white`, it really looks better.

project/src/com/google/daggerqueryui/scripts/highlight-query.js
o-y: I don't believe there's any specific jQuery style guides to observe. Google has their own JS style guide - https://google.github.io/styleguide/jsguide.html, you could take this one step further and integrate [Google's ESLint config](https://github.com/google/eslint-config-google) to automatically format your JS code and provide suggestions.

LGTM

Ratchette (approved):

project/src/com/google/daggerqueryui/scripts/highlight-query.js
margaritiko: Thank you!

PR #29, opened 2020-09-01
Simplifies Java code for j2cl compatibility

margaritiko:
Simplifies Java code for j2cl compatibility.

**We decided to use `j2cl` to reuse queries execution logic.**

**Problems and solutions:**
* `String.format(...)` method [isn't supported by `j2cl`. ](https://gwt.googlesource.com/gwt/+/master/user/super/com/google/gwt/emul/java/lang/String.java#73) All uses of this method have been replaced with simple string concatenation.
* `Apache Commons Text` isn't supported by `j2cl`. Implemented a method which calculates the distance between two strings using dynamic programming [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) algorithm.
* `j2cl protobuf` isn't fully supported. I have contacted `j2cl` team, and they told me that `j2cl_proto` works in Google3, but it's not ready yet to work with Bazel. I have tried to use `new_j2cl_proto_library` but it complains about dependencies ([full output](https://gist.github.com/margaritiko/98c33d7d81d7c30abaf486f728c5cfd8) for [BUILD file](https://gist.github.com/margaritiko/21325e7d75d866fb1cb806b305f2df9f)) what could be caused by a bug in j2cl protobuf implementation. The solution is to convert `BindingGraph` proto model to a simple adjacency list representation `Map<String, List<String>>`. This approach also greatly simplifies the creation of fake graphs in tests and improves the readability of the query implementation (albeit with additional memory costs).
* `NotImplementedException` has been replaced with `UnsupportedOperationException`.

### DIFF
--- a/project/src/com/google/daggerquery/executor/QueryExecutor.java
+++ b/project/src/com/google/daggerquery/executor/QueryExecutor.java
@@ -21,10 +21,15 @@ import com.google.daggerquery.executor.models.MisspelledNodeNameException;
 import com.google.daggerquery.executor.models.Query;
 import com.google.daggerquery.executor.services.SourcesLoader;
 import com.google.daggerquery.protobuf.autogen.BindingGraphProto.BindingGraph;
+import com.google.daggerquery.protobuf.autogen.DependencyProto.Dependency;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


margaritiko:
Close this PR because I have found another approach to deal with `j2cl protobuf` problem.

PR #30, opened 2020-09-01
Supports j2cl in query_executor_models target

margaritiko:
Supports j2cl in the target with queries implementation.

**Main changes:**
* Supported `j2cl` by adding required calls into WORKSPACE file.
* Supported j2cl `Guava`.
* Marked with `@JsType` annotation `Query` and `MisspelledNodeNameException` classes as we would like to use them in JavaScript code.
* Created new target with `j2cl_library` rule which can be used in js targets in `deps`.


project/repository.bzl
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Bazel rule for loading external repository deps for j2cl_proto."""
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+_JSINTEROP_BASE_VERSION = "master"
```
kwchan1995: I think it would be preferable to pin down the version for reproducible builds and avoiding surprise breakages if master introduces backwards incompatible change.

project/WORKSPACE
```diff
@@ -54,3 +54,76 @@ http_jar (
     name = "apache-commons-text",
     url = "https://repo1.maven.org/maven2/org/apache/commons/commons-text/1.9/commons-text-1.9.jar"
 )
+
+http_archive(
+    name = "com_google_j2cl",
+    strip_prefix = "j2cl-master",
```
kwchan1995: ditto, depending on `master` directly could be risky.

project/WORKSPACE
```diff
@@ -54,3 +54,76 @@ http_jar (
...
+load("@com_google_j2cl//build_defs:repository.bzl", "load_j2cl_repo_deps")
+load_j2cl_repo_deps()
+
+load("@com_google_j2cl//build_defs:rules.bzl", "setup_j2cl_workspace", "j2cl_maven_import_external")
+setup_j2cl_workspace()
+
+http_archive(
+    name = "com_google_elemental2",
+    strip_prefix = "elemental2-master",
+    url = "https://github.com/google/elemental2/archive/master.zip",
```
kwchan1995: ditto

margaritiko: I have checked [version 1.1.0](https://github.com/google/elemental2/releases) and it works, let's use that one.

project/WORKSPACE
margaritiko: Hm, I have failed to find any version [here](https://github.com/google/j2cl) which I can link to - there is only one release **v20190918**, but it's not full and doesn't contain `j2cl_maven_import_external` for Guava import 😶

project/repository.bzl
margaritiko: Yes, that's important -[version 1.0.0](https://github.com/google/jsinterop-base/releases/tag/1.0.0) works pretty nice, I have changed it ✅

project/WORKSPACE
kwchan1995: I see, never mind then, thanks for checking!

kwchan1995 (approved):

Ratchette (approved):

PR #31, opened 2020-09-01
j2cl compatibility support

margaritiko:
Simplifies Java code for j2cl compatibility.

**We decided to use `j2cl` to reuse queries execution logic.**

**Problems and solutions:**
* `String.format(...)` method [isn't supported by `j2cl`. ](https://gwt.googlesource.com/gwt/+/master/user/super/com/google/gwt/emul/java/lang/String.java#73) All uses of this method have been replaced with simple string concatenation.
* `NotImplementedException` has been replaced with `UnsupportedOperationException`.
* Library `j2cl protobuf` isn't fully supported. I have contacted `j2cl` team, and they told me that `j2cl_proto` works in Google3, but it's not ready yet to work with Bazel. I have tried to use `new_j2cl_proto_library` but it complains about dependencies ([full output](https://gist.github.com/margaritiko/98c33d7d81d7c30abaf486f728c5cfd8) for [BUILD file](https://gist.github.com/margaritiko/21325e7d75d866fb1cb806b305f2df9f)) what could be caused by a bug in j2cl protobuf implementation. The solution is to abstract from the proto models and use the interface `Graph` which provides an access to the `BindingGraph`.


project/src/com/google/daggerquery/executor/models/Graph.java
```diff
@@ -0,0 +1,44 @@
...
+ * The interface representing the graph created by the Dagger SPI plugin.
+ *
+ * <p>Provides the methods needed to traverse this graph and to execute queries.
+ */
+public interface Graph {
+
+  /**
+   * Returns all node's dependencies.
+   */
+  List<String> getDependencies(String node);
```
kwchan1995: nit: Would it be possible to return `ImmutableList`? This makes it clear the graph is readonly.

project/src/com/google/daggerquery/executor/models/Graph.java
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * Checks if a node is presented in the graph or not.
+   */
+  Boolean containsNode(String node);
+
+  /**
+   * Returns all the nodes presented in the graph.
+   */
+  Set<String> getAllNodes();
```
kwchan1995: ditto, would it be possible to return `ImmutableSet`?

project/src/com/google/daggerquery/executor/models/Graph.java
kwchan1995: or `ImmutableSet`

project/src/com/google/daggerquery/executor/models/Graph.java
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * Returns all node's dependencies.
+   */
+  List<String> getDependencies(String node);
+
+  /**
+   * Checks if a node is presented in the graph or not.
+   */
+  Boolean containsNode(String node);
```
kwchan1995: why not `boolean`? will this method return `null`?

kwchan1995 (approved):

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -153,7 +149,7 @@ public Query(String typeName, String... parameters) {
       }
     }
 
-    throw new NotImplementedException();
+    throw new UnsupportedOperationException("Query with specified name is not supported yet.");
```
Ratchette: Would you mind including the query name in this string?

Ratchette (approved):
LGTM 💐

project/src/com/google/daggerquery/executor/models/Graph.java
margaritiko: Yes, let's use `ImmutableSet`, because ordering doesn't make sense here and it will make it faster to check for a connection between nodes!

project/src/com/google/daggerquery/executor/models/Graph.java
margaritiko: 👌

margaritiko: pokemon

margaritiko: Done 😎

project/src/com/google/daggerquery/executor/models/Graph.java
margaritiko: I have changed to `boolean`, I completely forgot about `null`, sorry :)

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: Yep, I have added it ⭐️

PR #32, opened 2020-09-01
Implements Levenshtein distance algorithm 🦀

margaritiko:
Implements Levenshtein distance algorithm.

In order to use `j2cl` we need to get rid of using Apache Commons Text library because it isn't supported. Implemented a method which calculates the distance between two strings using dynamic programming [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) algorithm.

### DIFF
--- a/project/src/com/google/daggerquery/executor/models/Query.java
+++ b/project/src/com/google/daggerquery/executor/models/Query.java
@@ -28,7 +28,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
-import org.apache.commons.text.similarity.LevenshteinDistance;
 import sun.reflect.generics.reflectiveObjects.NotImplementedException;
 
 import static java.util.stream.Collectors.toList;
@@ -48,11 +47,6 @@ public class Query {
   private final static String SOMEPATH_QUERY_NAME = "somepath";
 
   private final static int MAX_NUMBER_OF_MISPLACED_LETTERS = 3;
-
-  /**
-   * An algorithm for measuring distances between node names in a binding graph to detect typos.
-   */
-  private final static LevenshteinDistance levenshteinDistanceCalculator = new LevenshteinDistance();
 
   /**
    * The key is the name of supported query and the value is the number of parameters,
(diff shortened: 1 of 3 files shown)


project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -308,12 +302,50 @@ private void checkNodeForCorrectness(String node, BindingGraph bindingGraph) {
...
+      for (int secondIndex = 0; secondIndex < secondLength; secondIndex++) {
+        if (firstString.charAt(firstIndex) == secondString.charAt(secondIndex)) {
+          distances[firstIndex + 1][secondIndex + 1] = distances[firstIndex][secondIndex];
+        } else {
+          int replaceCost = distances[firstIndex][secondIndex] + 1;
+          int insertCost = distances[firstIndex][secondIndex + 1] + 1;
+          int deleteCost = distances[firstIndex + 1][secondIndex] + 1;
+
+          distances[firstIndex + 1][secondIndex + 1] =
+              Arrays.asList(replaceCost, insertCost, deleteCost).stream().min(Integer::compareTo).get();
```
kwchan1995: Can be simplified with [`Collections.min`](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#min-java.util.Collection-) :smiley:

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -308,12 +302,50 @@ private void checkNodeForCorrectness(String node, BindingGraph bindingGraph) {
...
       if (distance <= MAX_NUMBER_OF_MISPLACED_LETTERS) {
         closestNodes.add(node);
       }
     }
 
     return closestNodes;
   }
+
+  /**
+   * Measures the difference between two strings.
```
kwchan1995: nit: Make it clear it's the Levenshtein distance in the Javadoc?

kwchan1995 (approved):

project/src/com/google/daggerquery/executor/models/Query.java
```diff
@@ -308,12 +302,50 @@ private void checkNodeForCorrectness(String node, BindingGraph bindingGraph) {
...
+   *
+   * @return the number of changes required to get one string from another,
+   * where each change is a single character modification (substitution, deletion or insertion)
+   */
+  private int calculateDistance(String firstString, String secondString) {
```
Ratchette, line 322: Should we make this algorithm case insensitive?

Ratchette (approved):

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: I'm not sure - since nodes names are case sensitive, it makes sense to distinguish `com.google.A` from `com.google.a`.

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: 👌

project/src/com/google/daggerquery/executor/models/Query.java
margaritiko: 😄😄😄✔️

project/src/com/google/daggerquery/executor/models/Query.java
Ratchette: You're right :)

PR #35, opened 2020-09-09
Split the main method in QueryExecutor into two different 🧩

margaritiko:
Refactor `QueryExecutor` class and split the `main` method into two different.

To reuse this logic, which takes an array with the name of the request and its parameters, let's create a separate method.

**Main changes:**
* method `main(String[] args)` is responsible for taking command line arguments, requesting the result of the query from another method, and printing it to the console
* method `execute(String[] args)` executes the query and puts the results into a list.


kwchan1995 (approved):

Ratchette (approved):
LGTM ⛰

PR #36, opened 2020-09-09
Implemented Java server

margaritiko:
Implement a server which gets requests with queries and executes them.

**Main changes:**
* Supported [Undertow](https://github.com/undertow-io/undertow) - a Java web server. Extended `WORKSPACE` with the required calls and placed the `BUILD` file in the appropriate third-party subfolder.
* Implemented `Server` class which starts the server and manages requests.
* Created `BUILD` file with `java_library` which contains `Server` class and dependencies.

**Invariant:** query execution method always returns paths, `deps` query is not an exception: its paths always contain only one node and therefore nothing is separated via ‘->’.

We convert all results into `JSON` array via [Gson library](https://github.com/google/gson) and then send them as a string (https://undertow.io/javadoc/1.3.x/io/undertow/io/Sender.html).

**Why port equals 4921?** Well, let's take "Dagger Query" string, [get its sha256](https://xorbin.com/tools/sha256-hash-calculator), take first four digits and reverse them 🙃

[Card](https://github.com/googleinterns/dagger-query/projects/2#card-44227014)


project/src/com/google/daggerquery/server/Server.java
```diff
@@ -0,0 +1,73 @@
...
+  private static void executeQuery(HttpServerExchange exchange) {
+    exchange.getResponseHeaders().put(new HttpString("Access-Control-Allow-Origin"), "*");
+    exchange.getResponseHeaders().put(new HttpString("Access-Control-Allow-Methods"), "GET");
+
+    Deque<String> deque = exchange.getQueryParameters().get("query");
+    if (deque == null) {
+      return;
+    }
+
+    String[] args = deque.stream().filter(value -> value.length() > 0).toArray(String[]::new);
```
margaritiko: I am not sure about this line: I mentioned that `getQueryParameters().get("query")` returns a deque which always contains one more element - just an empty string. My request looks like this: `http://localhost:4921/daggerquery/?query=deps&query=com.google.Beach`.

It is not documented behavior. In this line, I am filtering all empty strings from the array before passing them to the `QueryExecutor`.

project/src/com/google/daggerquery/server/Server.java
```diff
@@ -0,0 +1,73 @@
...
+import io.undertow.util.HttpString;
+import io.undertow.util.StatusCodes;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Deque;
+import java.util.stream.Collectors;
+
+public class Server {
+
+  private static final int port = 4921;
```
o-y: As these are immutable constants I'd use UPPER_SNAKE_CASE. http://go/java-style#s5.2.4-constant-names

margaritiko: Thank! I completely forgot to rename them.
Done ✅

project/src/com/google/daggerquery/server/Server.java
kwchan1995: I'm not sure if the order of repeated params are guaranteed to be preserved, if not this design could be fragile, say the URL ends up as `query=com.google.Beach&query=deps`. How about taking the query as one single argument, e.g. `query=deps%20com.google.Beach`?

project/src/com/google/daggerquery/server/Server.java
```diff
@@ -0,0 +1,73 @@
...
+      return;
+    }
+
+    String[] args = deque.stream().filter(value -> value.length() > 0).toArray(String[]::new);
+    try {
+      ImmutableList<String> results = QueryExecutor.execute(args);
+      ByteBuffer[] buffers = results
+        .stream()
+        .map(result -> ByteBuffer.wrap(result.getBytes()))
+        .toArray(ByteBuffer[]::new);
```
kwchan1995: I'm not sure if you'd have difficulty decoding this from the client side. If it works then great. If not, maybe you could try packaging this as a JSON array (e.g. [use gson to convert Java to JSON](https://stackoverflow.com/a/36634293), [set content type to application/json](https://www.stubbornjava.com/posts/handling-content-types-with-undertow#application-json))

project/src/com/google/daggerquery/server/Server.java
margaritiko: O, great solution, thanks. It also will help us to get rid of these filtering calls because we can just take the first element from a deque.
Done ✅

project/src/com/google/daggerquery/server/Server.java
margaritiko: I just checked and everything is decoded correctly, **but** as I can see, it reads the content of the response as one line and the elements are not separated by any character. So we cannot understand where the first binding name ends, it is a problem(

In this case JSON can help. But as I see now [there is no special method](https://undertow.io/javadoc/1.3.x/io/undertow/io/Sender.html) for sending JSONs - or we can pack it as a `ByteBuffer` object again?

kwchan1995: `new Gson().toJson(...)` returns a `String`, which you can just pass on to `exchange.getResponseSender().send(...)`

margaritiko: Exactly, sorry for this clarification. I did it, it looks very simple and much better 🎀

kwchan1995 (approved):

project/src/com/google/daggerquery/server/Server.java
```diff
@@ -0,0 +1,72 @@
...
+
+      exchange.setStatusCode(StatusCodes.OK);
+      exchange.getResponseHeaders().put(Headers.CONTENT_TYPE, "application/json");
+      exchange.getResponseSender().send(new Gson().toJson(results));
+    } catch (IllegalArgumentException e) {
+      exchange.setStatusCode(StatusCodes.BAD_REQUEST);
+      exchange.getResponseSender().send("Execution failed. Reason: " + e.getMessage());
+    } catch (IOException e) {
+      exchange.setStatusCode(StatusCodes.NOT_FOUND);
+      exchange.getResponseSender().send("File with binding graph sources not found. Reason: " + e.getMessage());
```
Ratchette: This sounds like a problem with the server setup, that isn't actionable from the UI. 

If we were setting up a public webserver (eg. something that was designed for more than localhost) I'd recommend that we send back an internal server error status code to the UI, then print this error message to the server's stdout. 

... But, our tool is for devs only. So I don't think this is a big deal. Feel free to implement or ignore this advice as you like.

Ratchette (approved):

Ratchette:
Would you mind adding tests for the server here or in a follow-up CL?

margaritiko:
> Would you mind adding tests for the server here or in a follow-up CL?

I prefer to add it here, since PR with features and without tests is not a good idea 😼

margaritiko:
I [have added a test](https://github.com/googleinterns/dagger-query/pull/36/commits/3f2c53d5016b4a3c79264dd77ac1c812e3e049ac) which checks that server is launched correctly and returns `404` status code when cannot find a graph. We cannot mock binding graph, and we need to create a separate target with Java classes marked with Dagger annotations to test it in real, so let's add additional tests in next PRs.

project/src/com/google/daggerquery/server/Server.java
margaritiko: Thank you. Yes, since it for devs only maybe it makes sense to show this problem in UI too, but I will also add an [output](https://github.com/googleinterns/dagger-query/pull/36/commits/a27e8f1157f4c82b302194804136b844497f53e1) to the server's output stream.

project/tests/com/google/daggerquery/server/ServerTest.java
```diff
@@ -0,0 +1,74 @@
...
+  @BeforeClass
+  public static void startServer() {
+    // We can pass null since these arguments are not used.
+    Server.main(/*args =*/ null);
+
+    uriBuilder = new URIBuilder()
+      .setScheme("http")
+      .setHost("localhost")
+      .setPort(4921)
+      .setPath("/daggerquery/");
```
Ratchette: Would it be possible to re-use the static final variables from the server class, maybe by making them public static final?

Ratchette (approved):

project/tests/com/google/daggerquery/server/ServerTest.java
margaritiko: Yep, [done](https://github.com/googleinterns/dagger-query/pull/36/commits/3cd296eac7b615908109897c271e1f064c673afc) ✅

PR #37, opened 2020-09-10
Added bazel rule for creating a server

margaritiko:
Add bazel rule for creating a server and connecting target with resources to it.

**How to use?**
```
dagger_query_server(
    name = "java_server",
    dagger_app_target = "//src/com/google/daggerquery/example:example" # Name of the main app target
)
```

[Card](https://github.com/googleinterns/dagger-query/projects/2#card-44227014)

### DIFF
--- a/project/src/com/google/daggerquery/server/dagger_query_server.bzl
+++ b/project/src/com/google/daggerquery/server/dagger_query_server.bzl
@@ -16,8 +16,10 @@ load("@rules_java//java:defs.bzl", "java_binary")
 load("//src/com/google/daggerquery/plugin:dagger_query_textproto.bzl", "dagger_query_textproto")
 
 def dagger_query_server(name, dagger_app_target):
+   binding_graph_data_name = name + "_binding_graph_data"
+
    dagger_query_textproto(
-       name = "binding_graph_data",
+       name = binding_graph_data_name,
        dagger_app_target = dagger_app_target
    )
 
@@ -25,5 +27,5 @@ def dagger_query_server(name, dagger_app_target):
      name = name,
      main_class = "com.google.daggerquery.server.Server",
      runtime_deps = ["//src/com/google/daggerquery/server:server"],
-     resources = [":binding_graph_data"],
+     resources = [":" + binding_graph_data_name],
    )
--- a/project/src/com/google/daggerquery/server/dagger_query_server.bzl
+++ b/project/src/com/google/daggerquery/server/dagger_query_server.bzl
@@ -1,0 +1,29 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_java//java:defs.bzl", "java_binary")
+load("//src/com/google/daggerquery/plugin:dagger_query_textproto.bzl", "dagger_query_textproto")
+
+def dagger_query_server(name, dagger_app_target):
+   dagger_query_textproto(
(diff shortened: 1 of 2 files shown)


project/src/com/google/daggerquery/server/dagger_query_server.bzl
```diff
@@ -0,0 +1,29 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_java//java:defs.bzl", "java_binary")
+load("//src/com/google/daggerquery/plugin:dagger_query_textproto.bzl", "dagger_query_textproto")
+
+def dagger_query_server(name, dagger_app_target):
+   dagger_query_textproto(
+       name = "binding_graph_data",
```
kwchan1995: It would be better to change this to `name = name + "_binding_graph_data"` to avoid collisions if multiple `dagger_query_server` targets are added by the user in the same `BUILD` file.

project/src/com/google/daggerquery/server/dagger_query_server.bzl
```diff
@@ -0,0 +1,29 @@
...
+       dagger_app_target = dagger_app_target
+   )
+
+   java_binary(
+     name = name,
+     main_class = "com.google.daggerquery.server.Server",
+     runtime_deps = ["//src/com/google/daggerquery/server:server"],
+     resources = [":binding_graph_data"],
```
kwchan1995: Don't forget to update this to `":" + name + "_binding_graph_data"` to point to the new target. Maybe create a variable `binding_graph_data_name = name + "binding_graph_data"` so you could avoid the duplication.

kwchan1995 (approved):
Nice!

project/src/com/google/daggerquery/server/dagger_query_server.bzl
margaritiko: It will cause a problem since we have [a specific name in SourcesLoader](https://github.com/googleinterns/dagger-query/blob/26f380ba6d2c989c3ae01430b36f25afa36155e7/project/src/com/google/daggerquery/executor/services/SourcesLoader.java#L36). 

It can be solved by hardcoding "binding_graph_data.zip" name as an output [here](https://github.com/googleinterns/dagger-query/blob/26f380ba6d2c989c3ae01430b36f25afa36155e7/project/src/com/google/daggerquery/plugin/dagger_query_textproto.bzl#L42). Since it is not very important for us that the name of the archive and the target coincide, we can do this, right?

kwchan1995: The `binding_graph_data.zip` shouldn't cause a problem, because it is scoped to the runfiles directory of current Bazel invocation. The concern I was referring to is just the target name, because it will be resolved to `//<user's BUILD file>:binding_graph_data`. If the user's `BUILD` file declared multiple `dagger_query_server` targets, the Bazel system will complain about this target being declared more than once, when it expands the `dagger_query_server` macros in the user's `BUILD` file.

margaritiko: Hm, I mean that the name which we specify in `dagger_query_textproto` will be the name of .zip file.

For example, 
```
dagger_query_textproto(
       name = "sample",
       dagger_app_target = dagger_app_target
)
```
will generate `sample.zip`.

kwchan1995: Oh sorry I missed that, you're absolutely right. I'm not sure if there could be a problem should the outputs collide. I can't think of a clean way to solve this off the top of my head, so I guess let's just go ahead and worry about it later...

margaritiko: [This approach](https://github.com/googleinterns/dagger-query/pull/37/commits/dab2d8b81254583907b07f0380d7e3f574978c7b) works. It's better than what we had before because the problem with multiple server targets has been solved ☀️Yep, we can think about something better later.

project/src/com/google/daggerquery/server/dagger_query_server.bzl
margaritiko: Sure, a new variable sounds pretty good. Done 🦀

kwchan1995 (approved):

Ratchette (approved):
LGTM. 

The usage instructions in PR's description were really helpful :)

PR #38, opened 2020-09-10
Create a script for sending requests to the server

margaritiko:
Create a script which analyses input field content and sends requests to the server.

Before sending requests **validates query** by checking its name and the number of parameters. When receives a response looks at its status: if it is `OK` then marks input field as valid. Otherwise if it is `BAD_REQUEST` or `NOT_FOUND` status then marks input field as invalid and shows an error message.

Requests are sent **asynchronously**.

front-end-requests


project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
```diff
@@ -0,0 +1,101 @@
...
+
+    request.onreadystatechange = function () {
+      if (request.status === 200) {
+        $(this).markInputFieldAsValid();
+      } else if (request.status === 404 || request.status === 400) {
+        $(this).markInputFieldAsInvalid(request.responseText);
+      }
+    }
+
+    const url = `http://localhost:4921/daggerquery/?${query.map(element => `query=${element}`).join('&')}`;
```
o-y: To simplify this code you could look into using the [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL)[1] class in JavaScript (I believe this internally uses [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)). Your logic here is rather simple, but definitely something to consider.

1: The MDN docs doesn't do a good job of demonstrating usage of the `URL` class, so [this](https://www.valentinog.com/blog/url/) might be a better resource.

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
```diff
@@ -0,0 +1,101 @@
...
+  if (!$(this).validateQueryName(queryName)) {
+    queryNameElement.hide();
+    return;
+  }
+
+  // Highlights query's name when it is valid.
+  queryNameElement.html(queryName).show();
+
+  if (event.key === 'Enter' && $(this).validateParameters(query)) {
+    const request = new XMLHttpRequest();
```
o-y: The modern replacement to `XMLHttpRequest` is the [fetch API](https://developers.google.com/web/updates/2015/03/introduction-to-fetch), this is better as it allows usage of the async/await and [promise](https://developers.google.com/web/fundamentals/getting-started/primers/promises) syntax.

Here's a few [examples](https://javascript.info/async-await) of how the await syntax could be implemented (if you want!)

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Marks input field as invalid and shows a container with error message.
+   * @param {string} errorMessage a error message which will be shown under the input field
+   */
+  $.fn.markInputFieldAsInvalid = function (errorMessage) {
+    const inputField = document.getElementById("query-input");
+    const errorContainer = document.getElementById("error-message");
+
+    inputField.classList.remove("is-valid");
+    inputField.classList.add("is-invalid");
```
kwchan1995: Can be simplified to `$("#query-input").removeClass("is-valid").addClass("is-invalid")` (see https://api.jquery.com/removeclass/)

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
kwchan1995: +1, much simpler and more readable

kwchan1995: Alternatively, jquery supports this as well (https://api.jquery.com/jQuery.get/)

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
```diff
@@ -0,0 +1,103 @@
...
+    errorContainer.classList.remove("hidden");
+    errorContainer.innerHTML = errorMessage;
```
o-y: IIUC there's a risk of an XSS injection attack here (however one caused by and only affecting the user).

From the [gif](https://user-images.githubusercontent.com/19249980/92711385-2bc61300-f361-11ea-9682-8687c2840182.gif) you provided it looks like the input is directly relayed back to the user from the HTTP request and set as HTML in the error-message div. A user could potentially enter HTML markup (or a script tag) which would then be rendered as HTML markup, due to the usage of the `innerHTML` attribute.

As you're expecting a text response from the server (and not HTML code) using `el.innerText` is the better option. If you plan on returning HTML in the future such as text encased in `<b>` tags, take a look into [goog.dom.safe](https://google.github.io/closure-library/api/goog.dom.safe.html), this would still require sanitising user provided input though.

See http://go/xss#dom-based-xss and http://go/xss-antipatterns#example-21-using-untrusted-input-to-set-innerhtml-in-javascript

o-y (changes requested):

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
kwchan1995: or use the jquery `.text(text)` setter (https://api.jquery.com/text/#text-text) which automatically escapes the contents

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
margaritiko: Yes, things look better and easier with the `URL` class, I replaced this simple string interpolation with your suggestion ✔️

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
margaritiko: Let's use the fetch API as it provides `then` and `catch` where we can directly call methods that change the color of the input field. Thank you 💫

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
margaritiko: That's very interesting, thanks. At this point `.text(...)` looks attractive because all the logic for representing errors can be simplified to a [single line](https://github.com/googleinterns/dagger-query/blob/3ce235a5311d3e3225bad54f6acabd1d71add380/project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js#L50). Done ✅

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
margaritiko: Thanks, now 2 lines instead of 6 😎😄

margaritiko:
> nit: I think `await` is redundant here as response.json() is not an `async` method

Hm, in all examples [here](https://javascript.info/fetch) and [here](https://www.carlrippon.com/fetch-with-async-await-and-typescript/) `await` is used before `.json(...)` call. 

Maybe it makes sense because of this point:
> The await keyword before a promise makes JavaScript wait until that promise settles.

margaritiko:
> try {
>   const graph = await getQueryResults(url);
>   $(this).markInputFieldAsValid();
> } catch (err) {
>   $(this).markInputFieldAsInvalid();
> }

`await` is only valid in `async` functions 🤔

Maybe, to deal with it I can create another function which will be `async`, will send requests, get responses and draw the entire graph? Or drawing from `async` function is not the best idea?

kwchan1995:
> > nit: I think `await` is redundant here as response.json() is not an `async` method
> 
> Hm, in all examples [here](https://javascript.info/fetch) and [here](https://www.carlrippon.com/fetch-with-async-await-and-typescript/) `await` is used before `.json(...)` call.
> 
> Maybe it makes sense because of this point:
> 
> > The await keyword before a promise makes JavaScript wait until that promise settles.

Sorry, you're right, my bad :disappointed:

kwchan1995:
> > try {
> > const graph = await getQueryResults(url);
> > $(this).markInputFieldAsValid();
> > } catch (err) {
> > $(this).markInputFieldAsInvalid();
> > }
> 
> `await` is only valid in `async` functions 
> 
> Maybe, to deal with it I can create another function which will be `async`, will send requests, get responses and draw the entire graph? Or drawing from `async` function is not the best idea?

Your suggestion makes sense. I don't see any issues drawing from within `async` functions.

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
```diff
@@ -0,0 +1,91 @@
...
+  if (event.key === 'Enter' && $(this).validateParameters(query)) {
+    async function getQueryResults(url) {
+      let response = await fetch(url);
+
+      if (response.status === 200) {
+        return await response.json();
+      }
+
+      throw await response.text();
+    }
```
o-y: At the moment you're redeclaring the same function (getQueryResults) each time a user triggers a change in the input container, you could consider moving it out of the keydown callback. (As this is a fairly generic util func, pulling it out gives scope for it to be used else where as well). The V8 engine is pretty well optimised for these kind of patterns so there would be a minimal performance hit, so your choice.

o-y (approved):
LGTM!

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
margaritiko: Interesting, thanks. We care about performance, so [let's move](https://github.com/googleinterns/dagger-query/pull/38/commits/d2564dd8785de1b82de46e0912d8841e43298d40) it to a new function (it'll be more readable as well).

kwchan1995 (approved):

Ratchette (approved):

PR #39, opened 2020-09-10
Remove j2cl support

margaritiko:
Remove j2cl support and required dependencies from WORKSPACE.

Since we decided to move forward with a **client-server approach**, we no longer need `j2cl`.


kwchan1995 (approved):

Ratchette (approved):
Thanks for cleaning this up! 🧹

PR #40, opened 2020-09-14
Support drawing all query results

margaritiko:
Support drawing all query results with [vis.js](https://visjs.org/).

**Main changes:**
* Implemented a `bindingGraph` instance that represents a subgraph of the initial binding graph that is displayed on the page. This graph is interactive and it's possible to add and remove nodes and edges. Let's assume that `n` **is the number of nodes in the subgraph** and `m` **is the number of edges**. Therefore, **time complexity** of the functions is the following:
    *  `addEdge` takes O(1) 
    * `deleteDeps` takes O(n^2) since for each target node we check to how many edges it is incident
    * `addPath` takes O(L) where `L` is the length of the given path
    * `clear`, `getNodes` and `getEdges` take O(1)
* Put vis.js file to the `third-party` library.

Since this PR contains many changes, tests that cover this functionality are located [in the following PR](https://github.com/googleinterns/dagger-query/pull/41).

(Sorry for quality, but video was quite heavy and after compression something went wrong 🙈)
Screen Recording 2020-09-14 at 20 06 56

[Card №1](https://github.com/googleinterns/dagger-query/projects/2#card-44227133)
[Card №2](https://github.com/googleinterns/dagger-query/projects/2#card-44227127)
[Card №3](https://github.com/googleinterns/dagger-query/projects/2#card-44227122)


project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
```diff
@@ -91,6 +91,20 @@ $(function () {
       url.searchParams.append('query', query.join(' '));
       const graph = await $(this).getQueryResults(url);
       $(this).markInputFieldAsValid();
+
+      if (query[0] === $.DEPS_QUERY_NAME) {
+        bindingGraph.clear();
+        for (let childNode of graph) {
```
kwchan1995: nit: since `childNode` is never mutated, you can use `const` instead of `let`.

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
```diff
@@ -91,6 +91,20 @@ $(function () {
...
+          bindingGraph.addEdge(query[1], childNode);
+        }
+      } else if (query[0] === $.ALLPATHS_QUERY_NAME || query[0] === $.SOMEPATH_QUERY_NAME) {
+        bindingGraph.clear();
+        for (let path of graph) {
```
kwchan1995: It seems like the structure of `graph` here (consisting of `paths`) is different from the one on line 97 (consisting of `nodes`). How about renaming `graph` to just `results` on line 92?

project/src/com/google/daggerqueryui/scripts/binding_graph.js
```diff
@@ -0,0 +1,212 @@
...
+  function getSimpleName(node) {
+    const firstIndex = node.indexOf('[');
+
+    if (firstIndex === -1) {
+      const components = node.split('.');
+      return components[components.length - 1];
+    }
+
+    const container = node.substring(0, firstIndex).split('.').slice(-1);
+    const content = getSimpleName(node.substring(firstIndex + 1, node.lastIndexOf(']')));
```
kwchan1995: What if the node name has multiple generic args, e.g. `com.google.Map[java.lang.String, com.google.Foo]`?

Alternative idea - how about doing a regex replace, so you don't have to worry about characters like `[`, `]`, `,`

```
const fullNameRegex = /(?:\w+\.)+(\w+)/g;
const simpleName = node.replaceAll(fullNameRegex, '$1');
return simpleName;
```

kwchan1995 (approved):
Awesome!

project/src/com/google/daggerqueryui/scripts/binding_graph.js
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * Tries to remove the given node from a subgraph.
+   *
+   * <p>The operation is considered successful only if the given node does not connect
+   * to any other node in the graph.
+   *
+   * @param {number} nodeId an identifier of a node to be removed
+   */
+  function tryToRemoveNode(nodeId) {
```
Ratchette, line 64: Could we return a boolean to indicate if the removal was successful?

project/src/com/google/daggerqueryui/scripts/binding_graph.js
```diff
@@ -0,0 +1,212 @@
...
+     * @param {string} target the second node of the given edge
+     */
+    addEdge: function (source, target) {
+      const sourceId = addNode(source);
+      const targetId = addNode(target);
+
+      edges.add({from: sourceId, to: targetId});
+    },
+
+    /**
```
Ratchette: It looks like this method isn't currently in use.
Is this functionality related to rdeps?

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
margaritiko: Thanks, that's right ✅

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
margaritiko: Yes, `graph` and `bindingGraph` names really add some ambiguity, let's rename it

project/src/com/google/daggerqueryui/scripts/binding_graph.js
margaritiko: Thanks, [done](https://github.com/googleinterns/dagger-query/pull/40/commits/dd646674d4fea935200d7c04825af61a19e7b0d6) 🎀

project/src/com/google/daggerqueryui/scripts/binding_graph.js
margaritiko: 😻 Thanks you, that's an awesome solution with the regex!

project/src/com/google/daggerqueryui/scripts/binding_graph.js
margaritiko: Yes, it will later be used to remove previously displayed deps. Since this PR contains quite a few changes, let's remove this feature, I will add it to a PR that supports click-to-remove 🐝

Ratchette (approved):
LGTM 🎨

PR #42, opened 2020-09-15
Add an example target

margaritiko:
Add an example target which was build with [Dagger](https://dagger.dev/) framework.

This target contains several files that simply describe the relationships between objects located in a vacation spot. The object diagram is below.

### To start a server with a connected target:
```
cd project
bazel run src/com/google/daggerquery:example_java_server
```

### Queries examples:
* **allpaths** com.google.daggerquery.example.HotelComponent com.google.daggerquery.example.models.CocaCola
* **somepath** com.google.daggerquery.example.HotelComponent com.google.daggerquery.example.models.Sprite 
* **deps** com.google.daggerquery.example.models.Villa

unnamed

_Note: Components are highlighted with yellow, containers which were created via `@Provides` annotation with pink and other nodes with blue._

### DIFF
--- a/project/src/com/google/daggerquery/example/models/Apartment.java
+++ b/project/src/com/google/daggerquery/example/models/Apartment.java
@@ -19,22 +19,18 @@ package com.google.daggerquery.example.models;
 import javax.inject.Inject;
 
 public class Apartment {
-  @Inject
-  Bed bed;
+  private final Bed bed;
+  private final MiniBar miniBar;
+  private final TV tv;
+  private final Phone phone;
+  private final Tourist tourist;
 
   @Inject
-  MiniBar miniBar;
-
-  @Inject
-  TV tv;
-
-  @Inject
-  Phone phone;
-
-  @Inject
-  Tourist tourist;
-
-  @Inject
-  Apartment() {
+  Apartment(Bed bed, MiniBar miniBar, TV tv, Phone phone, Tourist tourist) {
+    this.bed = bed;
+    this.miniBar = miniBar;
+    this.tv = tv;
+    this.phone = phone;
+    this.tourist = tourist;
   }
 }
--- a/project/src/com/google/daggerquery/example/models/Apartment.java
+++ b/project/src/com/google/daggerquery/example/models/Apartment.java
@@ -1,0 +1,40 @@
+/*
+Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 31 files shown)


project/src/com/google/daggerquery/example/models/Apartment.java
```diff
@@ -0,0 +1,40 @@
...
+  @Inject
+  TV tv;
+
+  @Inject
+  Phone phone;
+
+  @Inject
+  Tourist tourist;
+
+  @Inject
```
kwchan1995, line 28: You should pick either "constructor injection" or "field injection".

In the case of constructor injection, the constructor is marked with `@Inject`, but not the fields:

```
final MiniBar miniBar;
...

@Inject
Apartment(MiniBar miniBar, ...) {
  this.miniBar = minibar;
  ...
}
```

In the case of field injection, fields are marked with `@Inject`, but not the constructor:
```
@Inject MiniBar miniBar;
...

Apartment() {}
```

Constructor injection is typically preferred, because it allows the fields to be marked `final`.

project/src/com/google/daggerquery/example/models/Bed.java
```diff
@@ -0,0 +1,25 @@
...
+limitations under the License.
+ */
+
+package com.google.daggerquery.example.models;
+
+import javax.inject.Inject;
+
+public class Bed {
+  @Inject
+  public Bed() {
```
kwchan1995: nit: `@Inject` constructors should have default visibility instead of `public` to avoid users from manually invoking the `new` constructor. Dagger will generate factory code in the same package to invoke the constructor.

project/src/com/google/daggerquery/example/modules/AccommodationModule.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+@Module
+public class AccommodationModule {
+  @Provides
```
kwchan1995, line 32: I think you need `@ElementsIntoSet` to make this a `Set` multibinding collection. (see https://dagger.dev/dev-guide/multibindings.html#declaring-multibindings)

kwchan1995: Never mind, it shouldn't make any difference to the demo.

kwchan1995 (approved):
Nice demo!

Ratchette (approved):

project/src/com/google/daggerquery/example/models/Apartment.java
margaritiko: Thank you! I [have removed](https://github.com/googleinterns/dagger-query/pull/42/commits/b21367932cfc087466f3a202d08c7ec3f837cadc) all _field injection_ cases and made fields `private final`.

project/src/com/google/daggerquery/example/models/Bed.java
margaritiko: Yep, that's definitely makes sense, I [have removed](https://github.com/googleinterns/dagger-query/pull/42/commits/b21367932cfc087466f3a202d08c7ec3f837cadc) `public` keyword.

PR #43, opened 2020-09-16
Supports RDEPS query 🏔

margaritiko:
Supports `rdeps` query and covers implementation with tests.

Here is a trade-off between time and memory. We care about the time complexity of queries more so the solution is to save the reversed binding graph. That's will take us O(1) to get all ancestors of the source node.

This query was created to support showing all ancestors in Dagger Query UI. However, it will be available in the command-line app too and can be quite useful. 

**Main changes:**
* `rdeps` query implementation in the `Query` class
* creating the reversed binding graph in `GraphProto` constructor 
* extending `Graph` interface with method which return all ancestors
* adding tests in `QueryTest` class

### DIFF
--- a/project/src/com/google/daggerquery/executor/models/Graph.java
+++ b/project/src/com/google/daggerquery/executor/models/Graph.java
@@ -31,6 +31,11 @@ public interface Graph {
   ImmutableSet<String> getDependencies(String node);
 
   /**
+   * Returns all node's ancestors.
+   */
+  ImmutableSet<String> getAncestors(String node);
+
+  /**
    * Checks if a node is presented in the graph or not.
    */
   boolean containsNode(String node);
--- a/project/src/com/google/daggerquery/executor/models/Graph.java
+++ b/project/src/com/google/daggerquery/executor/models/Graph.java
@@ -31,7 +31,7 @@ public interface Graph {
   ImmutableSet<String> getDependencies(String node);
 
   /**
-   * Returns all node's ancestors.
+   * Returns all of a node's ancestors (nodes that depend on this node).
    */
   ImmutableSet<String> getAncestors(String node);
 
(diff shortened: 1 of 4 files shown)


project/src/com/google/daggerquery/executor/models/Graph.java
```diff
@@ -30,6 +30,11 @@
    */
   ImmutableSet<String> getDependencies(String node);
 
+  /**
+   * Returns all node's ancestors.
```
Ratchette: For some reason my brain didn't immediately associate "Ancestor" with "Something that depends on this node".
WDYT about the word dependant? eg. getDepentantNodes? (I'm not thrilled with this suggestion, since it sounds quite similar to the method above.)

Would you mind also mind elaborating in comment what an ancestor is?
eg. Returns all of a node's ancestors (nodes that depend on this node).

project/src/com/google/daggerquery/executor/models/GraphProto.java
```diff
@@ -29,9 +33,30 @@
...
+        reversedGraphModel.get(target).add(source);
+      }
+    }
+
+    this.reversedBindingGraph = ImmutableMap.copyOf(
+      Maps.transformValues(
+        reversedGraphModel,
+        ImmutableSet.Builder::build
+      )
+    );
```
kwchan1995: nit: How about moving this logic to a helper method for clarity?

kwchan1995 (approved):

Ratchette (approved):

project/src/com/google/daggerquery/executor/models/Graph.java
margaritiko: I chose this name because I thought it is quite common in graph theory, for example there is an LCA problem called [Least Common Ancestor](https://www.hackerrank.com/topics/lowest-common-ancestor), and so on.

margaritiko: I have clarified [what an ancestor is](https://github.com/googleinterns/dagger-query/pull/43/commits/c58f408b6947d79ca43ec11e4cc47f4830557c35) in this commit

project/src/com/google/daggerquery/executor/models/GraphProto.java
margaritiko: [Done](https://github.com/googleinterns/dagger-query/pull/43/commits/d2943267efcf2e2238b051c76b286acaad265bb2) ✅

PR #44, opened 2020-09-16
Saves all exceptions and prioritises them after execution 🛍

margaritiko:
Saves all exceptions in one set and prioritises them after all queries execution.

**Problem:** There are can be a situation when we have multiple graphs, and users type a query which result is empty (but source node exists). The result can depend on the order of out binding graphs because we remember only the last thrown `IllegalArgumentException`. And exception which indicates that the source node is missing in the last graph can overlap the previous exception, that no paths exist.

**Solution:** Let's save all raised exceptions in one set and then pick the first one. Each exception saved paired with an integer which represents a priority of exception (`1` means that exception is the most significant).

### DIFF
--- a/project/src/com/google/daggerquery/executor/QueryExecutor.java
+++ b/project/src/com/google/daggerquery/executor/QueryExecutor.java
@@ -17,6 +17,9 @@ limitations under the License.
 package com.google.daggerquery.executor;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Ordering;
+import com.google.common.collect.SortedSetMultimap;
+import com.google.common.collect.TreeMultimap;
 import com.google.daggerquery.executor.models.GraphProto;
 import com.google.daggerquery.executor.models.MisspelledNodeNameException;
 import com.google.daggerquery.executor.models.Query;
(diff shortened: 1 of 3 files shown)


project/src/com/google/daggerquery/executor/QueryExecutor.java
```diff
@@ -67,27 +72,22 @@ public static void main(String[] args) {
     ImmutableList.Builder<String> resultBuilder = new ImmutableList.Builder();
 
     // We assume that we successfully executed a query only if in at least one graph it was executed without fail.
-    MisspelledNodeNameException lastMisspelledNodeNameException = null;
-    IllegalArgumentException lastIllegalArgumentException = null;
+    Set<Entry<Integer, Exception>> exceptions = new TreeSet<>(Entry.comparingByKey());
```
kwchan1995: [`TreeMultimap`](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/TreeMultimap.html) is your friend.

kwchan1995 (approved):

Ratchette (approved):

project/src/com/google/daggerquery/executor/QueryExecutor.java
margaritiko: 😼😼😼
Thanks, [done](https://github.com/googleinterns/dagger-query/pull/44/commits/b8253b2e085e45ecdb327ad7fbe689d75f0a084a)!

PR #47, opened 2020-09-17
Supports clicks on the nodes

margaritiko:
Supports `right-click` and `double click` events in the network.

Let's use `double-click` event for showing dependencies. If at least one dependency has been already shown, then `double-click` on a node will hide all of them. Otherwise it will show all dependencies.

`Right-click` shows ancestors if they were not shown and hides all those present otherwise.

The graph redrawing logic is quite different for two modes: when **physics is disabled** all new nodes by default are placed at the one position and overlap. Therefore we need to redraw graph each time. 

static-3

With **enabled physics and non-zero gravity** nodes take different places after stabilisation is finished.

physics

[Card №1](https://github.com/googleinterns/dagger-query/projects/2#card-44227184)
[Card №2](https://github.com/googleinterns/dagger-query/projects/2#card-44227174)

### DIFF
--- a/project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
+++ b/project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
@@ -87,7 +87,7 @@ const queryExecutor = (function() {
      * @param {string[]} query a valid query which will be executed
      * @param {boolean} shouldClearGraph a flag which indicates if the graph should be cleaned or not
      */
-    processQuery: async function(query, shouldClearGraph) {
+    processQuery: async function(query, {shouldClearGraph}) {
       try {
         const url = new URL(`http://localhost:4921/daggerquery/`);
         url.searchParams.append('query', query.join(' '));
@@ -128,6 +128,6 @@ $("#query-input").on('keyup', function (event) {
   queryNameElement.html(queryName).show();
 
   if (event.key === 'Enter' && $(this).validateParameters(query)) {
-    queryExecutor.processQuery(query, true);
+    queryExecutor.processQuery(query, {shouldClearGraph: true});
   }
 });
--- a/project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
+++ b/project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
@@ -63,13 +63,13 @@ $(function () {
   };
 });
 
-$(function () {
+const queryExecutor = (function() {
   /**
    * Sends a request to the server with specified query and gets a response.
    * @param url an address with the correct path and request parameters in it
    * @return {Promise<any>} a promise that contains a json with a graph on success
    */
-  $.fn.getQueryResults = async function (url) {
+  async function getQueryResults(url) {
     let response = await fetch(url);
 
     if (response.status === 200) {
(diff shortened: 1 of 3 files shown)


project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
```diff
@@ -123,6 +128,6 @@ $("#query-input").on('keyup', function (event) {
   queryNameElement.html(queryName).show();
 
   if (event.key === 'Enter' && $(this).validateParameters(query)) {
-    $(this).processQuery(query);
+    queryExecutor.processQuery(query, true);
```
o-y: nit: it's a bit unclear what functionality this boolean is introducing - it doesn't convey the meaning. You could add a parameter name comment. Ditto for all similar `processQuery` function calls

o-y (commented):
LGTM

Ratchette:
Would it be possible to add arrows to the edges to make it clearer which is the dependency?
(not necessarily in this PR)

project/src/com/google/daggerqueryui/scripts/binding_graph.js
```diff
@@ -89,22 +100,40 @@ const bindingGraph = (function() {
...
    */
-  function getAllDeps(nodeId) {
+  function getAllEdgesToDeps(nodeId) {
     return edges.get({
       filter: edge => edge.from === nodeId
     });
   }
 
+  /**
+   * Retrieves all edges where the target node is the specified one.
```
Ratchette: nit: I'm not sure I understand what this means. Would you mind rephrasing it a little? eg.

Retrieves all edges where the target node is *a dependency*.

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
kwchan1995: Another option is to take advantage of [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Unpacking_fields_from_objects_passed_as_a_function_parameter) in javascript, so you could do:

```
queryExecutor.processQuery(query, {shouldClearGraph: true});
```
and in the definition

```
processQuery: async function(query, {shouldClearGraph}) {
   ...
}
```

project/src/com/google/daggerqueryui/scripts/binding_graph.js
```diff
@@ -138,6 +167,50 @@ const bindingGraph = (function() {
...
+      // Checks if any node was selected.
+      if (params.nodes.length === 0) {
+        return;
+      }
+
+      const nodeId = params.nodes[0];
+      const nodeTitle = nodes.get(nodeId).title;
+
+      // If the dependencies are not currently shown, we prefer to draw them all.
+      // Otherwise, we hide all children, even if they are not all were presented.
```
kwchan1995: Thanks for the explanation! Would it be worth extracting the logic to a helper method, e.g. `showOrHideDeps`?

project/src/com/google/daggerqueryui/scripts/binding_graph.js
```diff
@@ -138,6 +167,50 @@ const bindingGraph = (function() {
...
+      if (getAllEdgesToDeps(nodeId).length === 0) {
+        queryExecutor.processQuery(['deps', nodeTitle], false);
+      } else {
+        bindingGraph.deleteDeps(nodeTitle);
+      }
+    });
+
+    // An event for managing parent nodes.
+    network.on("oncontext", function (params) {
```
Ratchette, line 219: It looks like from the vis documentation this is the event that is triggered when the user clicks with the right mouse button. https://visjs.github.io/vis-network/docs/network/

Will the chrome context menu also pop up when the user right clicks?

kwchan1995 (approved):

project/src/com/google/daggerqueryui/scripts/binding_graph.js
```diff
@@ -138,6 +167,50 @@ const bindingGraph = (function() {
     }).length !== 0;
   }
 
+  /**
+   * Supports click and double click events in the network object.
+   * @param {vis.Network} network
+   */
+  function supportEventsRecognition(network) {
```
Ratchette, line 206: Are the two events currently:
1) Click twice with the left button for deps 
2) Click once with the right button for ancestors

If so, can we update each behaviour to be the same # of clicks or the same button? eg. 
1) Click once with the left button for deps 
2) Click once with the right button for ancestors

project/src/com/google/daggerqueryui/scripts/binding_graph.js
Ratchette: I see you've fixed this in the following PR :)
Resolving.

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
margaritiko: Thanks for your suggestions! [Let's use](https://github.com/googleinterns/dagger-query/pull/47/commits/a651a3fab549a43b37a23572991b1a38f9192522) the destructuring assignment syntax, it looks pretty clear.

margaritiko:
> Would it be possible to add arrows to the edges to make it clearer which is the dependency?
> (not necessarily in this PR)

Yep, sure! I plan to add it to PR with graph customization right after that.

project/src/com/google/daggerqueryui/scripts/binding_graph.js
margaritiko: Yes, good idea. I [have also moved](https://github.com/googleinterns/dagger-query/pull/47/commits/9faa992a1507a10c9bc6120f0c33a46dcdcae425) these comments to the JSDoc.

project/src/com/google/daggerqueryui/scripts/binding_graph.js
margaritiko: Yes, two events are exactly what you just described. At the beginning I though about using `left-click` and `double-click`, but it looks like `double-click` always raises two `left-click` events 😑. In vis.js docs they recommend to fix it by measuring time between clicks, but I don't like this approach since it is not really stable.

Instead, let's use `right-click`. I removed context menu [in this PR](https://github.com/googleinterns/dagger-query/pull/48) and it is easy to use it.

Now a single left-click action selects the node and its edges, but I don't think it really matters, so I don't see any problems using a single-click action instead of a double-click.

margaritiko: [Done](https://github.com/googleinterns/dagger-query/pull/47/commits/c6ffff60fb90ebae0f58b45c39b750977191b6c7)!

project/src/com/google/daggerqueryui/scripts/binding_graph.js
margaritiko: I mean, the node we passed to this method is equal to the destination node of the edge. I have paraphrased it [here](https://github.com/googleinterns/dagger-query/pull/47/commits/0b2d1e4ebabd8879e026e97769b41e9d8b02cd0e).

Ratchette (approved):

PR #49, opened 2020-09-18
Supported EXISTS query

margaritiko:
Supported **EXISTS** query which returns an array with a given source node if it exists and throws an exception otherwise.

For example, in graph: `com.google.A -> com.google.B -> com.google.C` query **exists com.google.C** will return `["com.google.C"]`, but query **exists com.google.D** will fail to execute (`MisspelledNodeNameException` will be thrown).


kwchan1995 (approved):
Nice!

Ratchette (approved):

PR #50, opened 2020-09-18
Supports RDEPS and EXISTS requests on the website side

margaritiko:
Supports RDEPS and EXISTS requests on the website side in `queryExecutor` module.

**Main changes:** 
* In `analyze_input_field_content.js `
   * Adds constants which represent RDEPS and EXISTS query
   * If the user presses `Enter`, the query name cannot be validated and the query contains one word, then if a node with that name exists, we draw it.
   * If parameters cannot be validated, marks an input field as invalid.
* In `bindingGraph.js`
   * Extends interface with functions which add ancestors / one node to the graph.

### DIFF
--- a/project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
+++ b/project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
@@ -41,10 +41,16 @@ $(function () {
    * Validates the number of query parameters.
    *
    * @param {Array<string>} query a full query passed by user with unique name and parameters
-   * @return {boolean} a value that determines whether the number of parameters is correct or not
+   * @throws a detailed error if number of parameters is wrong
    */
   $.fn.validateParameters = function (query) {
-    return $.supportedQueries.get(query[0]) === query.length - 1;
+    const queryName = query[0];
+    if ($.supportedQueries.get(queryName) === query.length - 1) {
+      return;
+    }
+
+    throw `The number of passed parameters for ${queryName} query is incorrect.` +
+    ` Expected: ${$.supportedQueries.get(queryName)}, got: ${query.length - 1}.`;
   };
 });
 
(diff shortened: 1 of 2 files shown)


project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
```diff
@@ -116,18 +124,32 @@ const queryExecutor = (function() {
...
 
-  if (event.key === 'Enter' && $(this).validateParameters(query)) {
-    queryExecutor.processQuery(query, true);
+  if (event.key === 'Enter') {
+    if ($(this).validateParameters(query)) {
+      queryExecutor.processQuery(query, true);
+    } else {
+      $(this).markInputFieldAsInvalid(
+        `The number of passed parameters for ${queryName} query is incorrect.` +
+        ` Expected: ${$.supportedQueries.get(queryName)}, got: ${query.length - 1}.`
```
kwchan1995: I'm worried this error message is fragile if the implementation of `validateParameters` is enhanced later. How about throwing this error from within the `validateParameters` function and catch it here, or change the function name to `validateNumParameters`?

project/src/com/google/daggerqueryui/scripts/binding_graph.js
```diff
@@ -238,6 +250,31 @@ const bindingGraph = (function() {
...
     },
 
+    /**
+     * Adds all nodes which depend on the source node.
+     *
+     * @param {string} source
+     * @param {string[]} ancestors
+     */
+    addAncestors: function (source, ancestors) {
+      for (const ancestorNode of ancestors) {
```
kwchan1995, line 277: Because of the `O(n)` `hasEdge` check within `addEdge`, the `addAncestors` operation now becomes `O(n^2)`. This may be fine for the demo though since the graph is small. If you run into performance issues later (laggy UI), you might want to optimise the `hasEdge` function, say keeping a copy of the edges in a separate `Set` (though this means 2x used memory as well).

kwchan1995 (approved):
Nice!

project/src/com/google/daggerqueryui/scripts/binding_graph.js
margaritiko: Yes, I agree with you, it's a problem. I am also not happy with calls like `getAllEdges`, `getAllEdgesToDeps` when they are used only for understanding it there are such edge or not.

In the next PR I will add adjacency list and save everything in it (it's better then set of edges since we can easily understand is there an edge which ends in the specified node, for example). I [have created a card ](https://github.com/googleinterns/dagger-query/projects/2#card-45756862)for this issue.

project/src/com/google/daggerqueryui/scripts/analyze_input_field_content.js
margaritiko: Let's [throw an exception](https://github.com/googleinterns/dagger-query/pull/50/commits/0729f1e537903c176f8a99207319220bd9239596) and hide the details in `validateParameters`.

✅

Ratchette (approved):

PR #51, opened 2020-09-18
Graph customisation

margaritiko:
Colours the graph depending on the request.

* When `deps` query is executed we draw source node with **green** and dependencies with **blue**.
Screenshot 2020-09-18 at 17 47 29

* When `allpaths` or `somepath` queries are executed we draw source and target nodes with **green**, others with **blue**.
Screenshot 2020-09-18 at 17 48 39

* When `exists` query is executed we draw a new node with **yellow**.

Screen Recording 2020-09-18 at 17 49 31


project/src/com/google/daggerqueryui/scripts/binding_graph.js
```diff
@@ -251,9 +287,12 @@ const bindingGraph = (function() {
...
     addPath: function (path) {
       const nodesInPath = path.split( ' -> ');
-      for (let index = 0; index + 1 < nodesInPath.length; ++index) {
-        addEdge(nodesInPath[index], nodesInPath[index + 1]);
+
+      addEdge(nodesInPath[0], nodesInPath[1], EdgeStyle.highlightSource);
+      for (let index = 1; index + 2 < nodesInPath.length; ++index) {
+        addEdge(nodesInPath[index], nodesInPath[index + 1], EdgeStyle.noHighlight);
       }
+      addEdge(nodesInPath[nodesInPath.length - 2], nodesInPath[nodesInPath.length - 1], EdgeStyle.highlightTarget);
```
kwchan1995, line 301: Is there a guarantee of path lengths `>= 2`? Otherwise might need to add a conditional check here to prevent index out of bounds (e.g. `allpaths Foo Foo`).

kwchan1995 (approved):

Ratchette:
This feature PR is a wonderful addition, thank you!

The initial draw looks great.
But it looks like the custom colouring becomes lost when the user manually traverses the graph (eg. left click & right click).
Is this intentional?

margaritiko:
> This feature PR is a wonderful addition, thank you!
> 
> The initial draw looks great.
> But it looks like the custom colouring becomes lost when the user manually traverses the graph (eg. left click & right click).
> Is this intentional?

Yes, at the moment it highlights only query results in different colors. I thought it might cause some performance issues if we need to redraw all the nodes for every click, but it would still be additional O (n), not so bad, so let's try adding some colouring while traversing. 

I think this feature will be an enhancement of current changes so I will add it in the next PR.

project/src/com/google/daggerqueryui/scripts/binding_graph.js
margaritiko: The realisation of `allpaths` and `somepath` queries in Dagger Query ensures that the path always contains at least the source and target nodes. But since the server can be changed and doesn't depend on the website, it is safer to add a conditional check.

[Done](https://github.com/googleinterns/dagger-query/pull/51/commits/2fe4c80d313c0319c3ea317cdac3e96aba00c71e) ✅

Ratchette (approved):

PR #52, opened 2020-09-18
Removes background image when input field is marked as valid or invalid

margaritiko:
Removes background image when input field is marked as valid or invalid.

**Problem:** when there are a lot of text in the input field it overlaps with an icon which appears after marking input field as valid or invalid. 

Screenshot 2020-09-18 at 23 07 30

Let's remove these icons.

Screenshot 2020-09-18 at 23 07 03


margaritiko:
There is also a problem where the highlighted query name name overlaps all the rest of the text, but that might be okay. Another solution which can be possible for highlighting a query name is to use `contentEditable` attribute with `div` tag. 

I think if it doesn't look good now, we can remove the highlighting entirely.

kwchan1995 (approved):

Ratchette (approved):

PR #56, opened 2020-09-21
Customise graph while traversing

margaritiko:
Repaints all nodes after changing the graph caused by clicking on the nodes.

**Strategy:** We have a queue with some colors that can be used to fill the nodes. Every time we want to recolor a node, we take the first color from this queue, use it and put it at the end of the queue. Also, before each execution of the query, we repaint all nodes in gray.

Screen Recording 2020-09-21 at 17 55 32


Ratchette (approved):

project/src/com/google/daggerqueryui/scripts/binding_graph.js
```diff
@@ -47,22 +47,92 @@ const bindingGraph = (function() {
...
+  const queueWithColours = ['#92cfff', '#fbaed2', '#93d145', '#f8cc43'];
+
+  /**
+   * Retrieves the first colour from the queue and places it at the end of the queue.
+   * @return {string} the first colour from the queue
+   */
+  function extractColor() {
+    const pickedColor = queueWithColours[0];
+    queueWithColours.push(pickedColor);
+    queueWithColours.shift();
```
kwchan1995: Since `.shift()` returns the first element, you can save one line :stuck_out_tongue: 
```js
const pickedColor = queueWithColours.shift();
queueWithColours.push(pickedColor);
```

project/src/com/google/daggerqueryui/scripts/binding_graph.js
```diff
@@ -47,22 +47,92 @@ const bindingGraph = (function() {
...
+   * Fills all nodes in the graph with the given color.
+   * @param {string} colour the colour that will fill all nodes
+   */
+  function recolourAllNodes(colour) {
+    nodes.forEach(node => nodes.update([{id: node.id, color: {
+        background: colour,
+        border: colour,
+        highlight: colour
+      }}])
+    );
```
kwchan1995: Pure speculation, maybe a single update call will be more efficient to reduce the number of events triggered

```js
const newColours = nodes.map(node => ({
  id: node.id,
  color: {
     background: colour,
     border: colour,
     highlight: colour
  }
}));
nodes.update(newColours);
```

kwchan1995 (approved):
very cool

project/src/com/google/daggerqueryui/scripts/binding_graph.js
margaritiko: 🚀 Thanks

project/src/com/google/daggerqueryui/scripts/binding_graph.js
margaritiko: Yes, it will really reduce the number of events calls significantly ✔️

PR #59, opened 2020-09-22
More instructions! 🎀

margaritiko:
Adds more instructions which shows how to use new queries, tool buttons and click actions.

Screenshot 2020-09-22 at 20 55 32


project/src/com/google/daggerqueryui/index.html
```diff
@@ -66,17 +66,33 @@
...
                         </li>
                     </ul>
+                    <p>
+                        To traverse an arbitrary graph use <b class="action-name">right-click</b> for showing node
+                        ancestors and <b class="action-name">left-click</b> for showing node dependencies.
+                        You can use the same actions to hide dependencies and ancestors as well.
+                    </p>
+                    <p>
+                        Consider the ability to simply <b class="action-name">specify the binding name</b>
+                        in the input field - if the graph contains such a node, it will be drawn on the plane.<br><br>
```
Ratchette, line 91: Nit:
Plane is a bit of an uncommon word, would you mind changing it here & the instance below to something like screen, page, or canvas?

Ratchette (approved):
Your documentation looks great. I love the colour coding :) 🍁

kwchan1995 (approved):
Love this!

project/src/com/google/daggerqueryui/index.html
margaritiko: Thank you! Let's replace it with **canvas**.

PR #60, opened 2020-09-23
Adds command for opening a website to the bash script

margaritiko:
Adds command for opening a website by executing the bash script.

User needs to pass the path to the server target. Otherwise, an example target will be executed as a default.


Ratchette (approved):
This is very helpful, thanks!

kwchan1995 (approved):

PR #64, opened 2020-09-23
Adds instructions for using Dagger Query UI to the README

margaritiko:
Adds instructions for using Dagger Query UI to the README.

In section **Dagger Query UI: Getting started** I put images which illustrates the result of the queries, and gif which shows how to traverse an arbitrary graph.


kwchan1995 (approved):

Ratchette (approved):
Your documentation is wonderful. 👩‍🚀

The images and gifs of the UI were super helpful

PR #68, opened 2020-09-30
Adds a solution for `permission denied` error

margaritiko:
Puts instructions in README which simply show how to give an **execute** and **read** permissions to the bash scripts.


Ratchette (approved):

README.md
```diff
@@ -26,6 +26,9 @@ Dagger Query can be used to investigate the dagger dependency graph.
     > ./dagger-query.sh YOUR_PATH allpaths com.google.Office com.google.Beach \
     > ./dagger-query.sh YOUR_PATH somepath com.google.Paris com.google.sights.EiffelTower
     
+   If permission is denied, run this command:
+    > chmod u+r+x dagger-query-ui.sh
```
kwchan1995: nit: `u+rx` is more idiomatic

kwchan1995: Actually you can make the file executable when you check it in to git (see https://stackoverflow.com/a/38285462), so that would save the user this step.

kwchan1995 (approved):

README.md
margaritiko: That's incredible, thank you!

I have added executable permissions to the scripts, but let's save this lines in README if on some platforms/machines it will still require do it manually.

kwchan1995: Sounds good!

PR #69, opened 2020-09-30
Removes directories from example project and improves README ✨

margaritiko:
Removes directories from example project and improves README.

**Main changes:** We decided to place an image that fully describes all the class relationships in the example project. In this graph we are only showing simple names, and it is not trivial to figure out what the real full name of the binding node is. Therefore, we will move all classes from the subdirectories `models` and `modules` to the `example` folder.


Ratchette (approved):
This is a nice improvement :)

kwchan1995 (approved):

PR #70, opened 2020-09-30
Highlights all paths instead of one

margaritiko:
Highlights all paths instead of one after executing `allpaths` query.

Since we have a public function which adds **one path** only it recolours the latest path which was shown. Let's create a public function which takes an **array with paths** and avoid recolouring all previous nodes with grey each time.

**Before:**
Screenshot 2020-09-30 at 21 41 28

**After:**
Screenshot 2020-09-30 at 21 39 28

### DIFF
--- a/project/src/com/google/daggerqueryui/scripts/binding-graph.js
+++ b/project/src/com/google/daggerqueryui/scripts/binding-graph.js
@@ -319,33 +319,50 @@ const bindingGraph = (function() {
     });
   }
 
-  return {
-    /**
-     * Takes a string representing a path where two adjacent nodes are connected by an edge,
-     * and adds all these edges to the subgraph.
-     *
-     * <p>Given path is separated with ' -> ' arrow.
-     *
-     * <p>If the path is invalid and contains less than two nodes, the method does nothing.
-     *
-     * @param {string} path a valid string representing a path with at least two nodes
-     */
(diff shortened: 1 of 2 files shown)


project/src/com/google/daggerqueryui/scripts/binding-graph.js
```diff
@@ -319,33 +319,50 @@ const bindingGraph = (function() {
...
+     * @param {string[]} paths an array with valid paths between nodes in the graph
      */
-    addPath: function (path) {
-      const nodesInPath = path.split( ' -> ');
-      if (nodesInPath.length < 2) {
-        return;
-      }
-
+    addPaths: function (paths) {
       recolourAllNodes(GRAY_COLOR);
```
Ratchette: Is this line still needed? It looks like you might already be setting the colour of all nodes in the addPath method.

Ratchette (approved):

kwchan1995 (approved):

project/src/com/google/daggerqueryui/scripts/binding-graph.js
margaritiko: Oh, yes, you are right. It is an unnecessary action, since the graph does not contain any nodes at this moment.
Removed ✅

PR #71, opened 2020-09-30
Adds commands for all platforms

margaritiko:
Adds commands which open a website for all platforms.

Since `open` command is available only on macOS, let's add `xdg-open` command which available in most Linux distributions and `start` command which works in Windows.

In case all three commands fail, there is an instruction in the README that teaches the user to open the website manually.


Ratchette (approved):

project/dagger-query-ui.sh
```diff
@@ -12,5 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+xdg-open src/com/google/daggerqueryui/index.html
+start src/com/google/daggerqueryui/index.html
```
kwchan1995: Well windows can't run `*.sh` scripts anyways so I don't think the `start` variant has value here. You might want to make available a separate `*.bat` script for windows if you really want :stuck_out_tongue:

project/dagger-query-ui.sh
```diff
@@ -12,5 +12,7 @@
...
 open src/com/google/daggerqueryui/index.html
```
kwchan1995: I think you want to conditionally execute `xdg-open` or `open` depending on the platform, otherwise either one of the commands will always fail and confuse the user.

You can do this with an if-else in bash, see https://stackoverflow.com/a/8597411

Alternatively just make available two different scripts, say:
* dagger-query-ui-linux.sh
* dagger-query-ui-macos.sh

margaritiko: Ok, using if-else in bash looks pretty straightforward, I [have added two checks](https://github.com/googleinterns/dagger-query/pull/71/commits/e4a939e9e592a5b8c450ed72aab61370b139ddf4): one for OSX, one for Linux. Thank you 🚀

project/dagger-query-ui.sh
margaritiko: O, makes sense. Let's just remove `start` command and mention in README, that Windows users need to run server manually (they still can open bash scripts and see, how to do it).

kwchan1995 (approved):

